PR #33, opened 2022-01-18
Innkeeper tenants

usingtechnology:


services/innkeeper/api/resources/tenant.py
```diff
@@ -0,0 +1,147 @@
...
+                "message": "Validation errors",
+                "errors": err.messages,
+            }, HTTPStatus.BAD_REQUEST
+
+        tenant = Tenant(**data)
+        if not Tenant.get_by_name(tenant.name):
+
+            # create a tenant in acapy - move to service
+            url = f"{current_app.config['ACAPY_ADMIN_URL']}/multitenancy/wallet"
+            headers = {
```
ianco, line 68: We'll want to refactor this at some point, but I think this is fine for the initial commit.

ianco (approved):
Overall looks really good!  Would be nice to have some updates to the README file though ...

services/innkeeper/api/resources/tenant.py
usingtechnology: yes, lots of refactoring and cleanup to do - LOTS.

PR #40, opened 2022-01-21
Feature/fastapi innkeeper

Jsyro:
migrating innkeeper functionaliy to use FastAPI, which endorses the use of sqlmodel, which harmonizes fastapi, sqlalchemy, and pydantic models. 

https://sqlmodel.tiangolo.com/

pagniation is not restored, but this might be an easy way to handle it.  https://uriyyo-fastapi-pagination.netlify.app/


services/traction/config.py
```diff
@@ -0,0 +1,28 @@
+import os
+
+
+class Config:
```
Jsyro, line 4: fastapi has a 'settings' module which has more support for loading from files and stuff.. might want to look into that too.

services/traction/requirements.txt
```diff
@@ -1,3 +1,33 @@
-uvicorn[standard]
-gunicorn
-fastapi
\ No newline at end of file
+alembic==1.7.5
```
Jsyro, line 1: this is just the output of `pip freeze` from within the container, might be more flexible ways of doing this.

Jsyro, line 1: this is just the output of `pip freeze` from within the container, might be more flexible ways of doing this.

PR #43, opened 2022-01-25
Aca-py utils for calling admin functions

ianco:
... plus a couple of example api's.

Work is WIP but opening PR for visibility.

Checks for tenant token in context before making call, so this PR will depend on the auth middleware for tenant calls to work.


services/traction/api/main.py
```diff
@@ -1,10 +1,62 @@
...
+
+from api.resources.ledger import router as ledger_router
 from api.resources.tenant import router as tenant_router
 
 from api.models import tenant  # noqa F401
+from api import acapy_utils as au
+
+
+
+class JWTTFetchingMiddleware(BaseHTTPMiddleware):
```
Jsyro, line 20: Fast is built on Starlette, https://fastapi.tiangolo.com/features/#starlette-features. So i'm wondering if these should fastapi references instead of starlette directly.

services/traction/api/resources/tenant.py
```diff
@@ -15,13 +17,28 @@
 router = APIRouter()
 
 
+class TenantWallet(BaseModel):
+    created_at: str
+    key_management_mode: str
+    settings: dict
+    state: Optional[str] = None
```
Jsyro, line 24: python 3.10 lets you just go `state: str | None = None` instead of needing the `typing.Optional`. 

https://www.python.org/dev/peps/pep-0604/

services/traction/migrations/versions/ad0e08a1c94f_tenant_table.py
```diff
@@ -1,7 +1,7 @@
 """tenant table
```
Jsyro, line 1: what wanted to re-format this file? it shouldn't matter, just odd that mine didn't do the same as yours.

Jsyro (commented):
Looks good, Just want wondering about the fastapi.middleware instead of directly importing the starlette stuff.

services/traction/migrations/versions/ad0e08a1c94f_tenant_table.py
ianco: It was just a whitespace issue.

```
$ PYTHONPATH=. flake8 .
./config.py:18:121: E501 line too long (129 > 120 characters)
./migrations/versions/ad0e08a1c94f_tenant_table.py:4:9: W291 trailing whitespace
```

ianco: Oh also black lol ...

```
$ black .
reformatted api/main.py
reformatted api/resources/ledger.py
reformatted migrations/versions/ad0e08a1c94f_tenant_table.py
reformatted migrations/env.py
reformatted api/acapy_utils.py
All done! ✨ 🍰 ✨
5 files reformatted, 6 files left unchanged.
```

services/traction/api/main.py
ianco: Possibly.  I'm just pulling in code from various examples at this point ...

PR #54, opened 2022-01-31
Add support for DIDExchange connections

ianco:


usingtechnology (approved):
👍 👍 
starting tomorrow (or very near future), we need to move this code to an underlying service and start using our own API for connections. Ran through these on the deployed PR, got connected to a BPA and to Trinisic Wallet Studio - awesome work!

PR #55, opened 2022-02-02
Update webhook security so that swagger docs/auth works

ianco:

Update webhook security to be more "FastAPI-ish"


usingtechnology (approved):
👍

PR #62, opened 2022-02-08
Create a demo session/sandbox with 3 tenants (Alice/Faber/Acme)

usingtechnology:

Had to re-work the Pydantic/SQLModel/SQLAlchemy models and repositories to handle ORM. I am not sure that the repository model will really work out; will have to rethink it. It is a real problem if the repository needs to return various different "read" schemas.

Anyway, no security anywhere. There is an empty `POST /webhook` for traction to call. We will need to implement and demonstrate the basic security mechanism and then actually handle the POSTed data.  But for now, it does get called by traction.

In this iteration, we can create new sandboxes (tag is optional but may help when we have a GUI just to know why you created this sandbox). End result of creating a sandbox is 3 new tenants in traction (Alice/Faber/Acme), that will be used to carry out the demonstration.


scripts/.env-example
```diff
@@ -114,6 +114,8 @@ ACAPY_ENDORSER_ENDPOINT=http://host.docker.internal:9030
 
 SHOWCASE_SERVICE_HOST=localhost
 SHOWCASE_SERVICE_PORT=5200
+SHOWCASE_ENDPOINT=http://host.docker.internal
+:5200
```
ianco, line 118: Line break?

scripts/docker-compose.showcase.yml
```diff
@@ -7,6 +7,7 @@ services:
     depends_on:
```
ianco, line 7: Wondering if it makes sense to combine everything into one docker-compose.yml?

ianco (changes requested):
Looks good!!!  One minor edit in the .env-example file and a question about the docker-compose file(s)

(Also update the README to describe how to start the showcase services)

scripts/.env-example
usingtechnology: whoops!

scripts/docker-compose.showcase.yml
usingtechnology: Let's discuss w/team. I think they should be separate because it's easier to fire up the "core" if I was using this to develop my own LOB against. We can achieve a lot with profiles... so whatever the rest of the team wants, I'm ok with. 6 of 1...

PR #71, opened 2022-02-10
Tweak to models and repositories.

usingtechnology:
Fix showcase tenant webhook path when registering


loneil (approved):
Think this looks good

PR #77, opened 2022-02-15
fix static files issues with docker compose.

usingtechnology:


ianco (approved):
Works like a charm!

PR #80, opened 2022-02-16
Upgrade aca-py to 0.7.3

ianco:

Updates one endpoint in the generated code.  It will be necessary to re-generate the generated wrapper and client at some point but nothing that affects us in the current sprint.

Also add the capability to connect using an agent's implicit invitation (public key) (we will use this to connect the tenant to the endorser)

### DIFF
--- a/scripts/docker-compose.yml
+++ b/scripts/docker-compose.yml
@@ -37,7 +37,7 @@ services:
       - ../services/traction:/app:rw
 
   traction-agent:
-    image: bcgovimages/aries-cloudagent:py36-1.16-1_0.7.2
+    image: bcgovimages/aries-cloudagent:py36-1.16-1_0.7.3
     depends_on:
       endorser-agent:
         condition: service_started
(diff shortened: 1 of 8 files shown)


scripts/docker-compose.yml
```diff
@@ -37,7 +37,7 @@ services:
       - ../services/traction:/app:rw
 
   traction-agent:
-    image: bcgovimages/aries-cloudagent:py36-1.16-1_0.7.2
+    image: bcgovimages/aries-cloudagent:py36-1.16-1_0.7.3
```
usingtechnology, line 40: Need to make the same change in the charts: endorser/values.yaml, traction/values.yaml

usingtechnology (changes requested):

scripts/docker-compose.yml
ianco: ok thanks!

usingtechnology (approved):

PR #81, opened 2022-02-17
acapy_wrapper: Fix URL to remove doubled "/" prefix

matgnt:
The "path" already comes with a "/" in the beginning.
No need to additionally add one. It even fails in e.g. /wallet/did


matgnt:
Makes sense. I'm now leaving default behavior as is and only set the url again, if path already starts with "/" which is the case (as you very correctly assumed) when using the wrapper standalone :-)

PR #91, opened 2022-02-23
Feature/showcase students

Jsyro:
Generate some random students along side Alice. Way more comfortable with python datageneration tools than doing this in JS. 

add explicit `_` after the name to indicate it was generated.


services/showcase/frontend/src/utils/constants.js
```diff
@@ -4,7 +4,7 @@
 
 /** API Route paths */
 export const ApiRoutes = Object.freeze({
-  BASEPATH: '/api/v1',
+  BASEPATH: 'http://localhost:5200/api/v1/',
```
Jsyro, line 7: Is there a known way to pass this via .env

services/showcase/api/main.py
```diff
@@ -46,7 +47,18 @@ def get_api() -> FastAPI:
...
+    "http://localhost:8080",
+]
+
+app.add_middleware(
+    CORSMiddleware,
+    allow_origins=origins,
+    allow_credentials=True,
+    allow_methods=["*"],
+    allow_headers=["*"],
+)
```
Jsyro, line 61: it's limited to localhost, so this is probably good to stay.

PR #97, opened 2022-02-28
Feature/showcase issuer

Jsyro:
Showcase automatically initiates the promotion to issuance step through the innkeeper API for the FABER tenant in new sandboxes.

New endpoint for the showcase tenant to complete the 'issuer registration' process.


services/showcase/api/db/migrations/versions/0e7a0e8f75ad_add_issuers_bools.py
```diff
@@ -0,0 +1,80 @@
...
+def upgrade():
+    # ### commands auto generated by Alembic - please adjust! ###
+    op.alter_column(
+        "outofband", "sender_id", existing_type=postgresql.UUID(), nullable=True
+    )
+    op.alter_column(
+        "outofband", "recipient_id", existing_type=postgresql.UUID(), nullable=True
+    )
+    op.alter_column(
+        "outofband", "sandbox_id", existing_type=postgresql.UUID(), nullable=True
```
Jsyro, line 29: I did not touch these columns, any idea why alembic wants to change from `nullable=false` to `nullable=true`?

usingtechnology: It is because you are generating based on the model class. The model class has those ids as optional because they are not part of the create schema (sandbox_id) or are to be optional on other schemas (update). you have to be very careful when generating migrations. (I don't bother, just code it by hand)

services/showcase/api/services/sandbox.py
```diff
@@ -230,3 +236,20 @@ async def accept_invitation(
...
+    sandbox_id: uuid.UUID,
+    tenant_id: uuid.UUID,
+    db: AsyncSession,
+) -> AcceptInvitationResponse:
+    sandbox = await get_sandbox(sandbox_id, db)
+    tenant = await get_tenant(sandbox, tenant_id, db)
+
+    # have tenant register itself as issuer
+    resp = await traction.tenant_admin_issuer(tenant.wallet_id, tenant.wallet_key, {})
+    logging.error(resp.status_code)
```
usingtechnology, line 251: logging/logger is not imported and declared

services/showcase/api/services/sandbox.py
```diff
@@ -230,3 +236,20 @@ async def accept_invitation(
         sender_id=sender.id,
         connection_id=uuid.UUID(resp["connection_id"]),
     )
+
+
+async def promote_tenant_to_issuer(
+    sandbox_id: uuid.UUID,
+    tenant_id: uuid.UUID,
+    db: AsyncSession,
+) -> AcceptInvitationResponse:
```
usingtechnology, line 245: Don't think this is the correct response type... is there one?

services/showcase/api/services/traction.py
```diff
@@ -199,3 +199,45 @@ async def accept_invitation(
...
+    async with ClientSession() as client_session:
+        async with await client_session.post(
+            url=t_urls.INNKEEPER_MAKE_ISSUER + f"/{tenant_id}",
+            headers=innkeeper_auth_headers,
+        ) as response:
+            try:
+                resp = await response.json()
+                return resp
+            except ContentTypeError:
+                logger.exception("Error accepting invitation", exc_info=True)
```
usingtechnology, line 217: Error accepting invitation isn't correct message

services/showcase/api/services/traction.py
```diff
@@ -199,3 +199,45 @@ async def accept_invitation(
...
+    async with ClientSession() as client_session:
+        async with await client_session.post(
+            url=t_urls.TENANT_MAKE_ISSUER,
+            headers=auth_headers,
+        ) as response:
+            try:
+                resp = await response.json()
+                return resp
+            except ContentTypeError:
+                logger.exception("Error accepting invitation", exc_info=True)
```
usingtechnology, line 238: Error accepting invitation isn't correct message

services/showcase/api/db/migrations/versions/0e7a0e8f75ad_add_issuers_bools.py
```diff
@@ -0,0 +1,80 @@
...
+# revision identifiers, used by Alembic.
+revision = "0e7a0e8f75ad"
+down_revision = "3449c639cdbd"
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    # ### commands auto generated by Alembic - please adjust! ###
+    op.alter_column(
```
usingtechnology, line 21: I would remove anything that you didn't explicitly want - I believe you just want the 3 new columns on tenant.

usingtechnology (changes requested):
call to make tenant an issuer will be done in swagger for the demo?

services/showcase/api/services/sandbox.py
Jsyro: i don't know why, but my `tox lint` is not catching these.

services/showcase/api/services/traction.py
Jsyro: fixed

services/showcase/api/services/sandbox.py
Jsyro: fixed, not sure if there is a better response

usingtechnology (approved):

PR #100, opened 2022-03-02
add a route for acapy admin api, secured.

usingtechnology:

endorser and traction expose acapy admin API routes (secured with admin API key secret ex. traction-acapy-api/adminurl).

added timeout to gunicorn to double from default of 30 seconds to 60. @ianco noticing that some of the workers were timing out and causing issues with workflows.  see if this helps.


ianco:
Should probably update the .env file so we run the admin api's secure locally as well ...

ianco (approved):
We can update the local .env later ...

ianco (approved):
Looks good!

PR #105, opened 2022-03-04
Presentation Exchange

ianco:

Adds workflow steps for presentation exchange (v10 protocol), and endpoints for verifier and holder/prover


services/traction/api/endpoints/routes/credentials.py
```diff
@@ -246,6 +265,125 @@ async def accept_credential(
...
+            CredPrecisForProof(
+                cred_info=cred.get("cred_info"),
+                interval=cred.get("interval"),
+                presentation_referents=cred.get("presentation_referents"),
+            )
+        )
+    return cred_precis
+
+
+@router.post("/holder/present-credential", response_model=PresentCredentialData)
```
usingtechnology, line 326: should this be `POST /holder/request` to match up with the verifier GET/POST? the models are the same

usingtechnology (approved):

PR #110, opened 2022-03-08
Tenant integration tests

ianco:


services/traction/conftest.py
```diff
@@ -31,7 +32,7 @@ async def db_session() -> AsyncSession:
         async with async_session(bind=connection) as session:
             yield session
             await session.flush()
-            await session.rollback()
+            await session.commit()
```
Jsyro, line 35: rolling back is what is supposed to keep each test isolated from the next. 

Shouldn't using the `app_client` fixture below give you the desired effect of committing the data changed by the test?

services/traction/conftest.py
```diff
@@ -66,5 +63,19 @@ def test_app(override_get_db: Callable) -> FastAPI:
...
 @pytest_asyncio.fixture()
 async def test_client(test_app: FastAPI) -> AsyncGenerator:
-    async with AsyncClient(app=test_app, base_url="http://test") as ac:
+    async with AsyncClient(app=test_app, base_url="http://localhost:5000") as ac:
+        yield ac
+
+
+@pytest.mark.integtest
+@pytest_asyncio.fixture()
+async def app_client() -> AsyncGenerator:
```
Jsyro, line 72: The suite of tests that require aca-py or multiple tenants appears to need this. It should probably be a 3rd test mark so they can be run selectively. i'll look into that later though.

ianco: Not sure why you want to run these tests "selectively", if we're running the integration test suite we should run all integration tests, otherwise we're making running the tests unnecessarily complicated.

services/traction/conftest.py
```diff
@@ -47,10 +48,6 @@ async def _override_get_db():
 @pytest.mark.integtest
```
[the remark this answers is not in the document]
ianco: Note sure what you mean here, I tried changing `@pytest.mark.integtest` to `@pytest.integtest` and the tests wouldn't run.

services/traction/conftest.py
Jsyro: There appear to be 3 test suites to me. 

1) Unit tests that only require python code (transient, no side effects)
2) Traction endpoint tests that require a db (session.rollback) 
3) Traction Integration tests that require webhooks and/or acapy (session.commit())

ianco: Option 2 is going to be a very limited set of tests.  The innkeeper tests (currently using the `test_client` fixture with rollback) all create a new tenant wallet, which doesn't get rolled back when the db gets rolled back.  I'm not sure what is the scope of tests that will only use the db.

Rolling back the db (but not the agent wallet(s)) is dangerous because the traction database will get out of sync with the agent, and it will still be possible for tests to interfere with each other, but will be more difficult to track down any resulting errors.

A better approach (imho, and also being used by the innkeeper tests) is to create new, randomized tenant(s) for each test, and then we can assure that tests won't interfere with each other.

PR #111, opened 2022-03-08
add more integration tests for innkeeper

usingtechnology:

### DIFF
--- a/services/traction/tests/integration/endpoints/routes/test_innkeeper.py
+++ b/services/traction/tests/integration/endpoints/routes/test_innkeeper.py
@@ -1,8 +1,16 @@
-import pytest
-import json
+import uuid
+from typing import List
 
-from tests.integration.db.tenant_factory import TenantCreateFactory
-from api.db.repositories.tenants import TenantsRepository
+import pytest
+
+from pydantic import parse_obj_as
+
+from api.db.models.tenant import TenantRead
+from api.db.models.tenant_issuer import TenantIssuerRead
+from api.endpoints.models.innkeeper import CheckInResponse
+from tests.integration.db.tenant_factory import (
+    CheckInRequestFactory,
+)
 from sqlalchemy.ext.asyncio import AsyncSession
 
 from httpx import AsyncClient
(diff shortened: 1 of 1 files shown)


services/traction/tests/test_utils.py
```diff
@@ -0,0 +1,51 @@
...
+async def innkeeper_auth(test_client: AsyncClient) -> str:
+    username = os.environ.get("TRACTION_API_ADMIN_USER", "innkeeper")
+    password = os.environ.get("TRACTION_API_ADMIN_KEY", "change-me")
+    data = {"username": username, "password": password}
+    token_resp = await test_client.post("/innkeeper/token", data=data)
+    assert token_resp.status_code == 200, token_resp.content
+    token_content = json.loads(token_resp.content)
+    bearer_token = "Bearer " + token_content["access_token"]
+
+    return bearer_token
```
Jsyro, line 23: Can these (innkeeper_auth and tenant_auth) be fixtures instead. That also means they can be scoped to 'module' so one auth token could get reused by multiple tests. 

If they aren't used by the test they don't run, so no downside.

services/traction/conftest.py
```diff
@@ -49,7 +49,7 @@ async def _override_get_db():
 def test_app(override_get_db: Callable) -> FastAPI:
     ##Disable Security before app is loaded
     # TODO will need to override again so we can test auth itself
-    s.ENDPOINT_SECURITY_ENABLED = False
+    s.ENDPOINT_SECURITY_ENABLED = True
```
Jsyro, line 52: If we want to run the tests this way, we should remove the infrastructure. This env_var was a hack.

Jsyro (changes requested):
Alot of these asserts should be separate tests. or at least parameterized. 
We could parameterize the expected outcome, (headers, json, expected_status_code).

https://docs.pytest.org/en/6.2.x/parametrize.html

services/traction/tests/test_utils.py
ianco: I don't think there's any benefit to making these fixtures, it makes the tests more opaque (not obvious that there is an auth process).  For tenants, we can't auth until we actually have a tenant, and we may have multiple tenants within a test.

services/traction/conftest.py
ianco: Yes agree this was just a patch to test out the authentication.

usingtechnology: leaving for ian's PR.

PR #112, opened 2022-03-09
feature/showcase-auto-accept-cred-offer

Jsyro:
showcase FE issues cred from faber-> alice
traction now bubbles custom event
showcase responds with the auto-accept call immediately. 


usingtechnology:
LGTM 👍

PR #113, opened 2022-03-10
Handle exception case for issue credential (reject offer)

ianco:


loneil (approved):
Looks good from what I can see. Good to get more tests in

PR #121, opened 2022-03-11
Feature/showcase notify ux

usingtechnology:


loneil (changes requested):
image
 discussed

loneil (approved):
End-to-end on Showcase functions looks like it's working

PR #122, opened 2022-03-11
ACME page - invite Alice

loneil:
Add ACME tab and list the applicants. 
Button to invite Alice applicant that sends invitation call. Display invitation status and other properties.

Has some quality-of-life UI things to sort out in the future (refresh things when certain actions happen etc) but will do for now for the invite functionality.


loneil:
https://pr-122-showcase-app-dev.apps.silver.devops.gov.bc.ca/

services/showcase/frontend/src/services/showcaseService.js
```diff
@@ -24,16 +24,28 @@ export default {
...
+   * @function createInvitationStudent
+   * Create a invitation for a student
    * @param {string} sandboxId The identifier for the sandbox
    * @param {string} lobId The identifier for the line of business
+   * @param {string} studentId The identifier for the student
    * @returns {Promise} An axios response
    */
-  createInvitation(sandboxId, lobId, studentId) {
+  createInvitationStudent(sandboxId, lobId, studentId) {
     return appAxios().post(`${ApiRoutes.SANDBOXES}/${sandboxId}/lobs/${lobId}/create-invitation/student`, { student_id: studentId });
```
Jsyro, line 35: Is this a generic method `createInvitation` or is this a faber specific method 'CreateInvitationtoAlice'... Just seems like it could be either and this is kind of a mix. 

Maybe just update the comments/local variables describing both roles. If someone passes in ACME's LobID to this method... there will be issues.. or at least 'undocumented behaviour'

Same for the method below. What is different between these two endpoints? Does the backend enforce the that the lobID has the intended role?

loneil: Yeah there's the 2 invitation type endpoints on the showcase API. I don't think the backend does LOB id checking or anything...
I'll rename the variables faberLobId and acmeLobId maybe?

usingtechnology (approved):
looks good.
maybe take note of @Jsyro comments and include those in the clean-up PR later today.

services/showcase/frontend/src/services/showcaseService.js
loneil: Have some changes planned to clean up this file upcoming in next batch of work, so will include this change with those in a next PR.

PR #123, opened 2022-03-11
Presentation exchange exception handling

ianco:
.... and adds an endpoint for the holder/prover to reject the presentation request

### DIFF
--- a/services/traction/api/endpoints/routes/credentials.py
+++ b/services/traction/api/endpoints/routes/credentials.py
@@ -444,6 +444,68 @@ async def holder_present_credential(
     return present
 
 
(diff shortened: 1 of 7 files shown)


services/traction/api/endpoints/routes/credentials.py
```diff
@@ -444,6 +444,68 @@ async def holder_present_credential(
     return present
 
 
+@router.post("/holder/reject-request", response_model=PresentCredentialData)
+async def holder_reject_presentation_requestl(
```
loneil, line 448: `l` at the end of the name intentional?

ianco: oops lol

loneil (approved):

PR #125, opened 2022-03-14
Feature/view cred offer

Jsyro:
Very crude ui, again showing the need for pretty labels and things. but the feature works.

image


ianco:
> * [ ]  that doesn't cause it to be removed from the set of offers.

The workflow should go to "error" state, so you should be able to filter by that (e.g. `params = {"state"="pending"}` will get you all the outstanding offers)

> * [ ]  also not sure if reject actually sends a problem report.

Check whether the state is updated, check the console logs ...?  How are you checking if the problem report is sent?

loneil (approved):
Looks good to me, did the invite/issue end-to-end and can see the Faber degree cred on Alice's page.
I've got some UI cleanup and data-refreshing flow changes going locally so I can prettify this a bit too.

PR #135, opened 2022-03-16
Feature/alice view proof request

Jsyro:
More crude FE and flow, but it works. a few rough edges I tried to document. 

I think traction should help the user isolate their tasks. e.g. presentation exchanges should be divided by role. A Presentation exchange you need to respond `role=holder` to is an entirely separate business item than a presentation exchange where `role=verifier`. 

broke lob endpoints in the showcase app into sub sections.


loneil (approved):
Did an end-to-end and functionality is working. I might move some things around UI-wise after.

PR #139, opened 2022-03-19
Add support for revocation for credential definition

ianco:
Adds options to require revocation support for a cred def
Adjust schema workflow to wait for at least one revocation registry
Update integration tests


loneil (approved):
Looks good. End-to-end regression through existing showcase cases working

PR #141, opened 2022-03-21
Store Schema Name to use instead of cred def.

usingtechnology:

Storing governance data in sandbox (updating as schema and cred def are created).
Pull out Schema Name instead of Cred Def ID in UX.
Use cred_def_id in sandbox.governance for presentation request (not Acme fetching from Faber's data).


loneil (approved):
Looks good, went through the end-to-end, see the degree name.

PR #149, opened 2022-03-25
Add revocation docs

ianco:

... also fixes one of the charts to include revocation notification params.


usingtechnology (approved):
👍 👍

PR #150, opened 2022-03-25
Feature/alice view proof request

Jsyro:
Update components for viewing cred offers and credentials. 

Not super consistent, but for now I think it allows up to better demonstrate the features. 

Could also make a similar change to presentation_requests  

image

image


loneil (approved):
End-to-end tested on deployed PR branch. Looks good, presentation flow working as expected

PR #152, opened 2022-03-25
Feature/pretty presentation request

Jsyro:
Changing view for presenations requests.

renames connection component to credential component as that's what it actually does. Currently no 'connection management' in alice 'portal'

image


loneil (approved):
Looking good, end to end working

PR #159, opened 2022-03-31
Show revocation message and notification

usingtechnology:

Screen Shot 2022-03-31 at 8 59 07 AM


loneil (approved):
Yeah looks good. Did the usual end-to-end and one with revocation. Looks like everything is displaying as expected in the showcase app.

PR #160, opened 2022-03-31
Update connection to use workflow

ianco:
Includes showcase updates

Note doesn't (yet) include support for the new traction connection notification


usingtechnology:
LGTM! 👍 👍

PR #163, opened 2022-04-06
V1 API RFC

usingtechnology:

Working on the URLs first, then fit in the response shapes, payload shapes and query parameters. 

Some notes:
- URLs that end with a trailing slash will denote collection. The reasons for the slash are: fast-api and 302 redirects require trailing slashes and it is easier to combine routes in an external router file that describes where it wants it mounted than having the root path in the resource file. 
- attempting to have all other URLs represent a final object or an action related to the collection.
- still need to address the more public URLs (unprotected URLs  - ex. tenant public profile)
- still need to address schemas and cred def listing and creation and where that should go.
- There are NO new innkeeper functions in this yet
- to see all the new endpoints, it is probably easiest to look at [tenant/docs#/v1](http://localhost:5100/tenant/docs#/v1)


ianco:
Just a general comment - the endpoints are all showing up on the swagger page under multiple url's, for example:

```
/v1/issuer/credential-definitions/
```

... shows up under `Credentials`, `v1` and `Issuer` headings.  I expect they all map to the same endpoint but the swagger is a bit confusing.

Also I think it's worthwhile separating out the endpoints that are *probably* not going to be integrated into an LOB (like `schemas`, `credential-definitions`, etc) - in `v0` these are under `admin` which I think makes more sense than just scattering them around everywhere.  That way a developer has a more clear idea of what they will have to integrate into their LOB vs not.

usingtechnology:
> Just a general comment - the endpoints are all showing up on the swagger page under multiple url's, for example:
> 
> ```
> /v1/issuer/credential-definitions/
> ```
> 
> ... shows up under `Credentials`, `v1` and `Issuer` headings. I expect they all map to the same endpoint but the swagger is a bit confusing.
> 
> Also I think it's worthwhile separating out the endpoints that are _probably_ not going to be integrated into an LOB (like `schemas`, `credential-definitions`, etc) - in `v0` these are under `admin` which I think makes more sense than just scattering them around everywhere. That way a developer has a more clear idea of what they will have to integrate into their LOB vs not.

yup, I just put them all under v1 so it's easier to see what we've added.  when we are done (or ready to ship alpha) we need to clean that up.  probably just slot all under v0, then put the v1 in appropriate sections.

usingtechnology:
was just for review, not needed any longer

PR #170, opened 2022-04-26
Issuer kit integration

ianco:
Changes required to work with issuer ket.


loneil (approved):
Looks good to me

services/showcase/api/endpoints/routes/webhooks.py
```diff
@@ -49,12 +48,7 @@ async def receive_webhook(
     db: AsyncSession = Depends(get_db),
 ):
     lob = await get_lob(lob_id, db)
-    try:
-        # request.json() is supposed to be json, but returns str :(
-        event_data = json.loads(await request.json())
-    except Exception:
-        # could not parse json
-        event_data = await request.body()
+    event_data = await request.json()
```
usingtechnology, line 51: 👍 👍

services/aca-py/Dockerfile.acapy
```diff
@@ -0,0 +1,10 @@
+FROM bcgovimages/aries-cloudagent:py36-1.16-1_0.7.4-rc0
```
usingtechnology, line 1: this is only for local development and running - correct? not to be deployed on openshift?

usingtechnology (approved):
A couple of questions, but looks good to me.

services/aca-py/Dockerfile.acapy
ianco: Ah right ...  I updated to 0.7.4-rc0 and there have been no issues, however we may want to wait for the "official" 0.7.4 release before we update the aca-py version on openshift.

FYI I expect `0.7.4-rc2` (I think is what we are up to) to come out this week, and then an official `0.7.4` next week.

With the official `0.7.4` we'll need to re-visit the aca-py client and wrapper libraries (openapi-generated code).  There are no differences in the new aca-py version for the api's that we're using, but there are changes to other api's and we should make sure we match ...

PR #171, opened 2022-05-03
Feature/v1 issuer api endpoints

Jsyro:
Issue uses contact_id not connection ID.

Schema's aren't issuer specific as we will want 'locally known' schemas for proof requests and proof request templates. 

holder still uses v0 endpoints. 

Not sure if v0 workflows should be visible or if we just use the 'status' value to manage the async workflow happening in the bg.


services/traction/api/db/models/v1/contact.py
```diff
@@ -83,6 +88,84 @@ class Contact(BaseModel, table=True):
         )
     )
 
+    @classmethod
```
usingtechnology, line 91: 👍 like this, let's keep doing this, simple gets/read functions in the "table" class. maybe we can add an optional bool for returning None or throwing the exception.

services/traction/api/endpoints/routes/v1/tenant/contacts/contact.py
```diff
@@ -33,7 +33,7 @@ async def get_contact(
     wallet_id = get_from_context("TENANT_WALLET_ID")
     tenant_id = get_from_context("TENANT_ID")
 
-    item = await contacts_service.get_contact(
+    db_item = await contacts_service.get_contact(
```
usingtechnology, line 36: this service call returns a ContactItem, the transformation in the route isn't needed. we will discuss, but these API services should be returning API level items not db

Jsyro: Yea i didn't got back and redo this after talked about it. 

Where do we want to put data operations? in the actual sqlmodel class? or do you see the need for 'Data Access Layer'

Internally we may want to update a contact as a side-effect, so I want to make sure that the 'update-contact' method isn't coupled with the endpoint. That's the only reason i thought we might want to do it this way. 

I'll make it consistent with the rest for now. my b

services/traction/api/endpoints/routes/v1/tenant/issuer/credentials.py
```diff
@@ -0,0 +1,141 @@
...
+from api.endpoints.models.v1.issuer import CredentialsListResponse, CredentialItem
+
+router = APIRouter()
+logger = logging.getLogger(__name__)
+
+
+@router.get("/credentials", response_model=CredentialsListResponse)
+async def get_issued_credentials(
+    state: TenantWorkflowStateType | None = None,
+    workflow_id: str | None = None,
```
usingtechnology, line 34: these workflow ids are internal, we aren't returning any internal data through the API (except where the id is actually used by LOB - ex. schema_id) or through the acapy data field that is explicitly asked for.

services/traction/api/endpoints/routes/v1/tenant/issuer/credentials.py
```diff
@@ -0,0 +1,141 @@
...
+    cred_issue_id: str | None = None,
+    db: AsyncSession = Depends(get_db),
+) -> List[IssueCredentialData]:
```
usingtechnology, line 37: this should be a ListResponse subclass as the return

services/traction/api/endpoints/models/v1/issuer.py
```diff
@@ -0,0 +1,37 @@
...
+class CredentialAcapy(BaseModel):
+    credential_exchange: dict = {}
+
+
+# class CredentialItem(
+#     AcapyItem[IssuerCredentialStatusType, ConnectionStateType, CredentialAcapy]
+# ):
+class CredentialItem(AcapyItem[str, str, CredentialAcapy]):  # v0
+    contact_id: Optional[UUID]  # v0 :UUID
+    alias: str
```
usingtechnology, line 31: what is alias in this context?

services/traction/api/endpoints/routes/v1/tenant/issuer/credentials.py
```diff
@@ -0,0 +1,141 @@
...
+    # this should take some query params, sorting and paging params...
+    wallet_id = get_from_context("TENANT_WALLET_ID")
+    tenant_id = get_from_context("TENANT_ID")
+
+    data = await issuer_service.get_issued_credentials(
+        db, tenant_id, wallet_id, workflow_id, cred_issue_id, state
+    )
+    logger.debug(data)
+    resp_data = [
```
usingtechnology, line 46: the service should be returning the API objects

services/traction/api/endpoints/routes/v1/tenant/issuer/credentials.py
```diff
@@ -0,0 +1,141 @@
...
+    cred_def_id: str | None = None,
+    contact_id: str | None = None,
+    alias: str | None = None,
+    db: AsyncSession = Depends(get_db),
+) -> CredentialItem:
+    wallet_id = get_from_context("TENANT_WALLET_ID")
+    tenant_id = get_from_context("TENANT_ID")
+
+    # Use connection ID for v0 compatability.
+    contact = await Contact.get_contact_by_connection_id(
```
usingtechnology, line 82: this is getting by contact id

services/traction/api/endpoints/routes/v1/tenant/issuer/issuer.py
```diff
@@ -0,0 +1,84 @@
...
+        # create workflow and update issuer record
+        tenant_workflow = await create_workflow(
+            wallet_id,
+            TenantWorkflowTypeType.issuer,
+            db,
+        )
+        # get updated issuer info (should have workflow id and connection_id)
+        tenant_issuer = await issuer_repo.get_by_wallet_id(wallet_id)
+
+    issuer = TenantIssuerData(
```
usingtechnology, line 55: we need new shapes that only contain relevant data to the LOB.

services/traction/api/endpoints/routes/v1/tenant/issuer/issuer.py
```diff
@@ -0,0 +1,84 @@
...
+        issuer=tenant_issuer,
+        workflow=tenant_workflow,
+    )
+
+    return issuer
+
+
+@router.get("/", status_code=status.HTTP_200_OK)
```
usingtechnology, line 63: get with "/" is a collection, not sure how we want to do this.... but looking at this endpoint i'd expect a list of issuers.

services/traction/api/endpoints/routes/v1/tenant/ledger/ledger.py
```diff
@@ -0,0 +1,101 @@
...
+    workflow: TenantWorkflowRead | None = None
+
+
+class TenantSchemaData(BaseModel):
+    schema_data: TenantSchemaRead | None = None
+    workflow: TenantWorkflowRead | None = None
+
+
+@router.get(
+    "/schema", status_code=status.HTTP_200_OK, response_model=List[TenantSchemaData]
```
usingtechnology, line 42: needs to be pluralized and end with "/", should return a newer List*Response

services/traction/api/endpoints/routes/v1/tenant/contacts/contact.py
usingtechnology: we will probably need a data access service at some point, just didn't want to go there until needed. i figure once we are worrying about cross table transactions we will need it.  i think for now, what you did for the fetch by various ids/fields in the sqlmodel is great. updates/deletes/searches can remain in the existing service layer. if we find we are recoding those exact updates/deletes/searches we can refactor out.

services/traction/api/endpoints/routes/v1/tenant/issuer/issuer.py
ianco: A tenant only has access to their own info, so this can only return a single item

services/traction/api/endpoints/routes/v1/tenant/issuer/credentials.py
Jsyro: removed that as an option for the v1 endpoint, good catch.

services/traction/api/endpoints/models/v1/issuer.py
Jsyro: It looks like it allows the caller to specify the recipient of the credential by the connection details instead of the contact_id. 

removing it for this endpoint.

services/traction/api/endpoints/routes/v1/tenant/issuer/credentials.py
Jsyro: That's not as feasible right now because the v0 endpoint uses that service method to (i refactored it that way). 

I could just fully duplicate the implementation then we can get this behaviour.

services/traction/api/endpoints/routes/v1/tenant/issuer/credentials.py
Jsyro: wow.... thanks

services/traction/api/endpoints/routes/v1/tenant/issuer/credentials.py
```diff
@@ -0,0 +1,141 @@
...
+    return response
+
+
+@router.post(
+    "/credentials/revoke",
+    status_code=status.HTTP_201_CREATED,
+    response_model=IssueCredentialData,
+)
+async def revoke_issued_credential(
+    cred_issue_id: str | None = None,
```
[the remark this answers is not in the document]
Jsyro: Yea, there should be a credential_id, but it looks like CredentialItem doesn't have that. good catch.

services/traction/api/endpoints/routes/v1/tenant/issuer/credentials.py
usingtechnology: the v1 service can wrap all the v0 stuff for now, that's what I was doing with contacts (create/receive an invitation), so in the future, you'd (theoretically) just change the internals of the service to do only v1 work.

services/traction/api/endpoints/routes/v1/tenant/issuer/issuer.py
Jsyro: I can drill down one level and just use the TenantIssuerRead class, but this is config/tenant level data, i'm not sure how to curate it as it's. 

If want to me rewrite that endpoint, i can. i just focused on using v0 code if it existed then tweaking the paths/payloads/responses.

services/traction/api/endpoints/routes/v1/tenant/issuer/issuer.py
Jsyro: could add a `/self` suffix if you want?

usingtechnology: i guess this would actually be better under a Tenant class? like what `v0/admin/tenant`, the information you want is the public did or if you are allowed to issue...  i kind of like adding a v1/self or something? GET would have that information in it, and we can hang off functions like make-issuer or any other thing like that...

services/traction/api/endpoints/routes/v1/tenant/issuer/credentials.py
```diff
@@ -0,0 +1,130 @@
...
+    )
+
+    return response
+
+
+@router.post("/", status_code=status.HTTP_201_CREATED, response_model=CredentialItem)
+async def issue_new_credential(
+    payload: IssueCredentialPayload,
+    db: AsyncSession = Depends(get_db),
+) -> CredentialItem:
```
usingtechnology, line 70: return should be based from GetResponse, so item = CredentialItem, links = [TBD]

usingtechnology (changes requested):
functionally i could get it all to work (except revocation, not sure where to get the rev_reg_id, cred_rev_id (but there is a comment around that anyway). need to discuss when things are getting to be v1 and more traction based and less acapy based.

services/showcase/api/endpoints/routes/line_of_business_issuer.py
```diff
@@ -50,7 +50,8 @@ async def issue_degree(
     # we will need some info for revoking this (rev_reg_id, cred_rev_id)
```
usingtechnology, line 50: i didn't think we were changing showcase... v0 was supposed to remain intact as is.
i guess this doesn't add or remove anything functionally, but we need to be careful here - was this just put in for debug purposes? if so, can you remove it, otherwise we have to redeploy showcase everywhere.

services/showcase/api/endpoints/routes/line_of_business_issuer.py
```diff
@@ -50,7 +50,8 @@ async def issue_degree(
...
     logger.info("issue credential response")
     logger.info(resp)
-    return resp
```
Jsyro, line 53: The issue was the internal showcase-api -> traction-api call was throwing a 404, and the showcase-api was returning that 404 as the body of a 200 response.

The the showcase FE successfully receives a response.. then the body of the response is actually an error message.

services/traction/api/endpoints/models/v1/governance.py
```diff
@@ -0,0 +1,36 @@
...
+    Attributes:
+      schema_request: required
+      schema_id:
+      cred_def_tag:
+      revocable:
+    """
+
+    # using existing schema_request shape from v0
+    schema_request: TenantSchemaRequest | None = None
+    schema_id: Optional[UUID]
```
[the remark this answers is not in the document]
Jsyro: done.

services/traction/api/endpoints/routes/v1/tenant/admin/issuer.py
```diff
@@ -0,0 +1,84 @@
...
+
+from api.db.models.tenant_workflow import TenantWorkflowRead
+from api.endpoints.models.tenant_workflow import TenantWorkflowTypeType
+
+router = APIRouter()
+logger = logging.getLogger(__name__)
+
+
+class TenantIssuerData(BaseModel):
+    issuer: TenantIssuerRead | None = None
```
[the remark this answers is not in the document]
Jsyro: made v1 for this endpoint.

services/traction/api/endpoints/routes/v1/tenant/issuer/credentials.py
Jsyro: good catch.

services/traction/api/endpoints/routes/v1/tenant/issuer/credentials.py
```diff
@@ -0,0 +1,130 @@
...
+        updated_at=data.workflow.updated_at,
+        alias="v0",  # alias is none, CredentialItem won't allow none
+        contact_id=payload.contact_id,  # v0
+    )
+
+    return response
+
+
+@router.post(
+    "/revoke",
```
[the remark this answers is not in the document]
Jsyro: we aren't using url parameters in other places... i don't oppose that being a url parameter, but i'm just worried it will be inconsistent. 

all of these are reasonable, not sure any of them are wrong. maybe there is a larger discussion there. 

`/contacts/<contact_id>/issue_credential` or 
`/issuer/credentials` or
`/governance/schemas/<schema_id>/issue`

services/showcase/api/endpoints/routes/line_of_business_issuer.py
Jsyro: see comment below, fixing a showcase bug.

services/traction/api/endpoints/routes/v1/tenant/issuer/credentials.py
```diff
@@ -0,0 +1,130 @@
...
+
+    data = await issuer_service.get_issued_credentials(
+        db, tenant_id, wallet_id, None, cred_issue_id, state
+    )
+    resp_data = [
+        CredentialItem(
+            **d.__dict__,
+            contact_id=d.credential.contact_id,
+            credential_id=d.credential.id,
+            status="v0",  # v0
```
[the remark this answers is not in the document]
Jsyro: We know there will be a `status` member to this model, we just don't know what it will be yet. so i thought this was a reasonable placeholder. 

Unless we want to discuss what it means? or is there going to be a global enum for 'ItemStatus' that we want to use.

services/showcase/api/endpoints/routes/line_of_business_issuer.py
usingtechnology: but this was working... so what really changed? does that mean the traction v0 changed? it's now throwing a different response in the same situation? if the fix needs to be in showcase, it needs to be done and we re-deploy it all but seems like there is new behaviour in v0 - when there shouldn't be any change.

services/traction/api/endpoints/models/v1/issuer.py
```diff
@@ -0,0 +1,91 @@
...
+    rev_reg_id: UUID
+    cred_rev_id: UUID
+    comment: str
+
+
+class CredentialAcapy(BaseModel):
+    credential_exchange: dict = {}
+
+
+class CredentialItem(AcapyItem[str, AcapyCredentialExchangeStateType, CredentialAcapy]):
```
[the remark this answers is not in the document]
usingtechnology: i added a ticket to the backlog, let's tackle it next sprint.

services/traction/api/endpoints/routes/v1/tenant/issuer/credentials.py
```diff
@@ -0,0 +1,130 @@
...
+) -> CredentialsListResponse:
+    # this should take some query params, sorting and paging params...
+    wallet_id = get_from_context("TENANT_WALLET_ID")
+    tenant_id = get_from_context("TENANT_ID")
+
+    data = await issuer_service.get_issued_credentials(
+        db, tenant_id, wallet_id, None, cred_issue_id, state
+    )
+    resp_data = [
+        CredentialItem(
```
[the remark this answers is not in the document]
usingtechnology: you write a v1 service that uses the v0 service for now... then you just change the internals of the v1 service when v0 is retired/removed.

services/traction/api/endpoints/routes/v1/tenant/issuer/credentials.py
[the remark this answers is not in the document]
usingtechnology: all 3 are valid, but you pick the one that makes the most sense... POST to the credentials collection makes sense, the you'd have the rest of the CRUD on the singular credential (ex. updating an external ref id, adding tags)... so you will end up with `GET /credentials/<id>`, `PUT /credentials/<id>` and then `POST /credentials/<id>/revoke` makes sense.

services/showcase/api/endpoints/routes/line_of_business_issuer.py
Jsyro: the comma is a `flask` short hand to set the status code of the response, but doesn't look like fastapi like that pattern anyways, reverted

usingtechnology: 👍

services/traction/api/endpoints/routes/v1/tenant/governance/schemas.py
```diff
@@ -0,0 +1,92 @@
...
+    schema_repo = TenantSchemasRepository(db_session=db)
+    tenant_schemas = await schema_repo.find_by_wallet_id(wallet_id)
+
+    response = SchemasListResponse(
+        items=tenant_schemas, count=len(tenant_schemas), total=len(tenant_schemas)
+    )
+    return response
+
+
+@router.post("/new", status_code=status.HTTP_200_OK)
```
usingtechnology, line 41: do not need "/new" here, just a post to the collection is fine and more consistent.  the "/import" is a special case.

usingtechnology (changes requested):
more little stuff... i know you are working on others so I'll stop testing/reviewing now until you say to go at it again.

services/traction/api/endpoints/routes/v1/tenant/admin/issuer.py
```diff
@@ -0,0 +1,57 @@
...
+        db,
+    )
+    # get updated issuer info (should have workflow id and connection_id)
+    tenant_issuer = await issuer_repo.get_by_wallet_id(wallet_id)
+
+    return tenant_issuer
+
+
+@router.get(
+    "/issuer/self", status_code=status.HTTP_200_OK, response_model=AdminTenantIssueRead
```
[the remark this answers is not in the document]
Jsyro: they are separate tables and repos, i could patch them together if we wanted to thought

usingtechnology (approved):

PR #172, opened 2022-05-04
Invitations - create reusable invitation. minimal API - list and create only

usingtechnology:
Add the most basic API for invitations, multi_use for now, add in public later.

Added handling of events for the invite to create a contact and add the tags to the new Contact.


loneil (approved):
Looks alright to me, ran through showcase use case end-to-end to regression test as well

services/traction/api/db/migrations/versions/v1_invitations_12a486c50ac3.py
```diff
@@ -0,0 +1,82 @@
...
+revision = "12a486c50ac3"
+down_revision = "0fbab19dd1f4"
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    # ### commands auto generated by Alembic - please adjust! ###
+    op.create_table(
+        "invitation",
```
Jsyro, line 23: do we want to be more specific? connection_invitation? 

Just wondering if we will have a different kind of invitation at some point.

usingtechnology: multi-use do not result in contact. we can move the other invitations here too if that makes sense. but this is quite a different use case. no resulting contact until someone uses this invitation, so we need to track this and the logic (future assumption is hanging a proof request workflow off this) differently.  we most certainly can move the create-invitation/receive-invitation to store data in this table. might make more sense.

Jsyro (approved):
Really great work. Liking the protocol listening structure.

services/traction/api/db/migrations/versions/v1_invitations_12a486c50ac3.py
usingtechnology: hmm, that is a very good point about different invitations, assuming you mean something like a traction invitation for tenancy?

PR #175, opened 2022-05-11
Feature/v1 issuer behave test

Jsyro:
should make some more 'composed' actions. 
 but cred offer and accept work.


loneil (commented):
We have a command for integration tests on this page
https://github.com/bcgov/traction/blob/develop/scripts/README.md
worth adding the step(s) for running the new Behave ones to that as well?

usingtechnology (approved):
👍 👍

loneil (approved):

PR #176, opened 2022-05-12
Bugfix/contact alias update

usingtechnology:
Bug on the reusable invitations processor, it was updating the alias for all contacts after the invitation was received. The logic is supposed to be for when a reusable invitation is received - as this is when we get the receiver's agent alias and we can set our "unknown" contact to be reasonably correct.

Updated the BDD tests that discovered the bug and worked around it.


Jsyro (approved):
Looks great!

PR #177, opened 2022-05-16
pass value from fetch to execute

Jsyro:


.vscode/settings.json
```diff
@@ -4,5 +4,8 @@
     "."
   ],
   "python.testing.unittestEnabled": false,
-  "python.testing.pytestEnabled": true
+  "python.testing.pytestEnabled": true,
+  "yaml.schemas": {
+    "https://json.schemastore.org/github-workflow.json": "file:///home/jsyro/src/digitrust/traction/.github/workflows/run_bdd_tests_dev.yml"
```
usingtechnology, line 9: don't know if we need this in the code base.

.github/workflows/run_bdd_tests_dev.yml
```diff
@@ -0,0 +1,40 @@
+name: Run Behvaiour Driven Development Tests
+on:
+  # pull_request:
```
usingtechnology, line 3: remove these comments

.github/workflows/run_bdd_tests_dev.yml
```diff
@@ -0,0 +1,40 @@
...
+  #   branches:
+  #     - develop
+  # schedule:
+  #   # * is a special character in YAML so you have to quote this string
+  #   - cron: "0 12 * * *"
+  workflow_dispatch:
+    inputs:
```
usingtechnology, line 10: looks like ref is the only parameter used right now.

services/traction/api/endpoints/routes/innkeeper.py
```diff
@@ -103,3 +103,19 @@ async def make_tenant_issuer(
         )
         tenant_issuer = await issuer_repo.create(new_issuer)
     return tenant_issuer
+
+
+@router.delete(
```
usingtechnology, line 108: are we leaving this as permanent? prefer just `DELETE /tenants/{tenant_id}?hard=true` or `DELETE /tenants/{tenant_id}?force=true`

services/traction/api/services/innkeeper.py
```diff
@@ -67,3 +70,21 @@ async def create_new_tenant(
...
+
+
+async def hard_delete_tenant(tenant: TenantRead, db: AsyncSession):
+    q = select(Tenant).where(Tenant.id == tenant.id)
+    result = await db.execute(q)
+    db_tenant = result.scalar_one()
+    if not tenant:
+        raise Exception
+
+    response = multitenancy_api.multitenancy_wallet_wallet_id_remove_post(
```
usingtechnology, line 82: this ends up using the tenant's wallet token correct? so it's a little weird to do the acapy work for the tenant in the innkeeper? could do the acapy work in tenant service, then call innkeeper to clean up the traction tables only.

services/traction/requirements-bdd.txt
```diff
@@ -0,0 +1,78 @@
+aiohttp==3.7.4.post0
```
usingtechnology, line 1: 👍 👍 keeping this separate from the traction service requirements!

usingtechnology (changes requested):
some stuff to clean up (gh action) and shuffle some code around.
awesome you got this working, huge step in the right direction.

.vscode/settings.json
Jsyro: oops

services/traction/api/endpoints/routes/innkeeper.py
Jsyro: i wanted it to be very distinct from a normal operational delete for now, borderline awkward.

if we do it 'correctly' then i think we can probably do something closer to what you have suggested here.

services/traction/api/services/innkeeper.py
Jsyro: I'm not sure how/why, but i did not end up needing the tenant to provide the acapy wallet_key. so the innkeeper endpoint with tenant_id is all you need

services/traction/requirements-bdd.txt
Jsyro: there is probably alot more that can be removed, but i'm not sure, so i'll revisit trimming that file down later

services/traction/api/services/innkeeper.py
usingtechnology: but i believe that since you are calling it via a tenant endpoint, the tenant's token gets put into the context which is what those acapy api clients use... so this is using the current context which is the tenant which is why it works.  so the innkeeper can't really call this function...

Jsyro: hmmmm, i'm calling from the innkeeper with the innkeeper auth token... so i'm not sure how acapy is using it's endpoint.

usingtechnology: sorry, my bad, yes, innkeeper route... i see. confused myself i guess...
so, inkeeper route, innkeeper service, innkeeper token, all good.

usingtechnology (approved):

loneil (approved):

PR #179, opened 2022-05-19
Default wallet scheme to MultiWalletSingleTableSharedPool

loneil:
For Traction we should be using the wallet scheme **MultiWalletSingleTableSharedPool** so that it's not creating databases for each tenant.
But allow this to be configurable by the Traction consumer.

Wallet schemes
https://github.com/hyperledger/indy-sdk/tree/master/experimental/plugins/postgres_storage#wallet-management-modes


usingtechnology (approved):
LGTM! 👍

PR #195, opened 2022-06-04
make v1 tenant model and update v1 tenant admin routes to use v1 responses

usingtechnology:
Still uses the v0 workflow.
Basically cosmetic changes for consistency in v1 api for right now.

Return a better response when the Tenant calls make-issuer before Innkeeper


loneil (approved):
Changes look good. Poked around in swagger, made tenant, tried both sides of making issuer in and out of order. Went through showcase flow.

PR #197, opened 2022-06-09
Track errors in background tasks, set Error status and message

usingtechnology:

Added error_status_detail to Issued Credential, Schema Template, Credential Template. These 3 classes are using (background) Tasks and we need to track when something goes wrong when we issue the call to AcaPy.  This is handled in the Task class and will set the status to Error and provide the error message in error_status_detail.

Moving forward, any further classes must have "Error" as a viable Status so that they will render properly in the API. 

Also, if the record has the status updated to something other than "Error" a trigger is fired to remove the error_status_detail.  So you need to put a trigger in the migration when adding error_status_detail.

Added more "base" models to reduce duplication, including the StatefulModel (Status, State, Error Status Detail) and using that for a new single Timeline table.


services/traction/api/db/models/base.py
```diff
@@ -36,3 +36,20 @@ class BaseTable(BaseModel):
...
+        sa_column=Column(TIMESTAMP, nullable=False, server_default=func.now())
+    )
+    updated_at: datetime = Field(
+        sa_column=Column(
+            TIMESTAMP, nullable=False, server_default=func.now(), onupdate=func.now()
+        )
+    )
+
+
+class StatefulModel(TimestampModel):
```
Jsyro, line 52: might be better to leave these separate (both inherit from `BaseModel`) Then we can pick and choose.

services/traction/api/db/models/v1/governance.py
```diff
@@ -21,14 +21,14 @@
...
-from api.db.models.base import BaseModel
+from api.db.models.base import BaseModel, StatefulModel
 
 from api.endpoints.models.v1.errors import (
     NotFoundError,
 )
 
 
-class SchemaTemplate(BaseModel, table=True):
+class SchemaTemplate(StatefulModel, table=True):
```
Jsyro, line 31: here you would just inherit from both `StatefulModel` and `TimestampModel`

services/traction/api/tasks/tasks.py
```diff
@@ -207,6 +216,27 @@ async def _perform_task(self, tenant: Tenant, payload: dict):
...
+            .values(values)
+        )
+        async with async_session() as db:
+            try:
+                await db.execute(q)
+            except DBAPIError:
+                await db.rollback()
+                self.logger.error(exc_info=1)
+            else:
+                await db.commit()
```
Jsyro, line 237: this chunk of code is gonna show up alot.... can this be a class method on the db class? just accepts a dict of {column:new_value}

we aren't really using the ORM at all, but this is a blind update, so i see the benefit.

Jsyro: It would also be great to see this generalized and centralised to the `StatefulModel` class... could use some reflection to find the pk column and use that id to update the row. I've written some reflection code I can point you to.

Jsyro (commented):
Great start, a few things i think can be improved

services/traction/api/tasks/tasks.py
usingtechnology: could do either way.

services/traction/api/db/models/v1/governance.py
usingtechnology: not sure we'd be having stateful model without timestamps? 🤷

services/traction/api/db/models/base.py
usingtechnology: did that.

services/traction/api/tasks/tasks.py
usingtechnology: set a utility function in base model, its an encapsulated read/update transaction/session.

services/traction/api/db/migrations/versions/v1_error_status_detail_8ac2e131039d.py
```diff
@@ -0,0 +1,223 @@
...
+
+create_error_status_func = """CREATE OR REPLACE FUNCTION error_status_detail_func() RETURNS trigger AS $body$
+    BEGIN
+        IF NEW.status <> 'Error' THEN
+            NEW.error_status_detail = NULL;
+        END IF;
+        RETURN NEW;
+    END;
+    $body$ LANGUAGE plpgsql
+"""
```
Jsyro, line 29: why do we need these as triggers? this should be enforced at the code level if this is critical.

usingtechnology: it's just easier to get all the cases done this way than enforce it in code. there is no single update function.

services/traction/api/db/models/base.py
```diff
@@ -36,3 +81,28 @@ class BaseTable(BaseModel):
...
+    )
+
+
+class StatefulModel(BaseModel):
+    status: str = Field(nullable=False)
+    state: str = Field(nullable=False)
+    error_status_detail: str = Field(nullable=True)
+
+
+class TimelineModel(BaseModel):
```
Jsyro, line 103: are these two related? can you have a TimelineModel without it also being also StatefulModel?

usingtechnology: They are different tables, Timeline is just tracking history of status it's strictly audit, not a business object table. you could add error_status_detail if wanted, but a lot of the status updates are just coming from acapy messages.

Jsyro: ah I see now, makes sense.

usingtechnology: we could still throw the error_status_detail in the timeline... may be good to see there what the error was, even if it gets corrected.

services/traction/api/db/models/base.py
```diff
@@ -36,3 +81,28 @@ class BaseTable(BaseModel):
...
+        sa_column=Column(
+            TIMESTAMP, nullable=False, server_default=func.now(), onupdate=func.now()
+        )
+    )
+
+
+class StatefulModel(BaseModel):
+    status: str = Field(nullable=False)
+    state: str = Field(nullable=False)
+    error_status_detail: str = Field(nullable=True)
```
Jsyro, line 100: https://docs.sqlalchemy.org/en/14/core/defaults.html

What about this. 

```
error_status_detail:: Field(
        sa_column=Column(
            str, nullable=False, onupdate=_secret_method
        )
    )
    
def _secret_method(context):
    if context.get_current_parameters()["state"] != "error":
       return None
    
    ```

File: services/traction/api/db/migrations/versions/v1_error_status_detail_8ac2e131039d.py (hunk 4 above)
Comment (Jsyro) [reply services/traction/api/db/migrations/versions/v1_error_status_detail_8ac2e131039d.py:29]: See comment in basemodel, i think there is a one line global fix.

File: services/traction/api/db/models/base.py (hunk 6 above)
Comment (usingtechnology) [reply services/traction/api/db/models/base.py:100]: if that works, let's put it in. and add it to timeline.  I'll take a look and get it in place.

Comment (usingtechnology) [reply services/traction/api/db/models/base.py]: couple of tweaks but works great, removed the trigger that was doing this work
https://github.com/bcgov/traction/blob/4efd739c443f878699774ef79f8633ad1fd75b32/services/traction/api/db/models/base.py#L100

File: services/traction/api/db/models/base.py [hunk 7]
```diff
@@ -36,3 +83,78 @@ class BaseTable(BaseModel):
...
+    )
+
+
+class StatefulModel(BaseModel):
+    def check_error_status_detail(context):
+        try:
+            if context.get_current_parameters()["status"] != "Error":
+                return None
+        except builtins.KeyError:
+            pass
```
Jsyro, line 105: I believe you need to return the value that will get saved, in either case. 

i think you need to return `context.get_current_parameters()["status"]` if status does == error. 

but i'm not sure of that, so if you tested it and it works, then that's great.

Jsyro (approved):
Great Work!

loneil (approved):

PR #200, opened 2022-06-16
Tenant UI - Simple Form

popkinj:
## Bare Bones Tenant UI
- Form to enter Wallet ID and Key
- Submit and clear buttons
- State to deactivate form during async request to API
- Requesting token from API
- Error handling failed requests
- Displaying token on screen if successful

## Development Environment
- Vite/Rollup stock development environment
- Typescript
- Docker config building production version of app
- Docker-compose integration with the rest of the project

## Actions
- I'll rebase after I open the PR
- Should I have the docker-compose port number as an Environment Variable?


services/traction/api/core/config.py
```diff
@@ -88,6 +88,9 @@ class GlobalConfig(BaseSettings):
 
     DEFAULT_PAGE_SIZE: int = os.environ.get("DEFAULT_PAGE_SIZE", 10)
 
+    # Allow CORS from a comma-separated list of origins
+    TRACTION_CORS_URLS: str = os.environ.get("TRACTION_CORS_URLS", "")
```
loneil, line 92: Think we should add this to env-example so people can just set it when they need it after copying over the example.
See how we do SHOWCASE_CORS_URLS in .env-example

loneil (changes requested):
Probably just an additon of that env var to the example file

services/traction/api/core/config.py
popkinj: Good thinking.

popkinj:
Tenant UI bar bones setup. Vue3, Vite & PrimeVue. Docker config incorporated into Docker-Compose.

PR #202, opened 2022-06-20
Store message content when permissions and configuration are enabled

usingtechnology:
Add BDD tests, include tests for tenant configuration and permissions.


loneil (approved):
I've set it up so that messages don't store

I have Innkeeper -> tenant permissions set to store_messages false
image
innkeeper/v1/tenants/3fab0576-de79-4e4b-835f-8bb4362d83af/permissions

Then on the Tenant side it still lets me set store_messages to true
image
tenant/v1/admin/configuration

Then I talk back and forth with a BC Wallet connection
image

And don't see any message content (null)
image

So that's expected. After turning it on on the innkeeper side the messages now store
image


Only 'concern' maybe is that in the second screenshot up there the tenant is still seeing that their config says to store messages, even though they don't cause the innkeeper permission is false. Accomplishing what we want but confusing maybe to the API user...? Not sure what a good API response solution would be though.

usingtechnology:
Yeha, that kind of bothers me too, so since you noticed it too let's deal with it. Probably the best course of action is to return an error (409?) with details that `store_*` require innkeeper set permissions first.

PR #203, opened 2022-06-22
Missing query resolver .all() on some list queries. 

Jsyro:
If you use the scalar result, it will resolve to the obj's JIT for some operations, but it is not the result, when returning that result, force resolution to list by using .all(). 

If not resolved properly, code will crash (expects list of data, but gets a query_result object). 


loneil (approved):
Tried some list endpoints and no errors I can see

PR #205, opened 2022-06-23
Feature/alembic model detection

Jsyro:
Alembic autogeneration wasn't synced up, now it gives no results, the only trick is to import the models into the __init__.py file so that when alembic runs, those models are actually loaded and linked properly.


loneil (approved):
Showcase tested end-to-end, did contact->issue on v1 to esatus regression as well

usingtechnology (approved):
👍

PR #206, opened 2022-06-23
Remove all unused event/notification code.

usingtechnology:

Just removing cruft from showcase and traction, prepping for an overhaul from using v0 to using v1 in showcase.


loneil (approved):
Did showcase regression and connect/issue to esatus and listing things in swagger. No issues

PR #208, opened 2022-06-30
Feature/showcase v0 to v1 issuer

usingtechnology:
Update make issuer and create schema and cred definition to v1.

Still makes use of old tenant table and old public did workflow.


usingtechnology:
I've run through Showcase with no issues.

loneil (approved):
Changes look good. Regression tested showcase and through swagger (new tenant, make issuer, V1 connection to a wallet)

PR #211, opened 2022-07-05
Feature/showcase v0 to v1 issuer credential

usingtechnology:
Showcase now offering/issuing and revoking credentials using v1.

There is a bit of wackiness without having holder functions in v1, and that is the tracking of credential to revoke. So we need to reach into the issuer tracking for now. This will only work if the issuer is in traction. There is a TODO in the code to address how we build the revocation notification once holder is fleshed out in v1.

Also, without going into a data modelling exercise, showcase is storing some v0 type data (cred def id, schema id, connection id) so we aren't really showing how best to use our v1 API.


loneil (approved):
Think it looks good. Tried Showcase flow end-to-end

PR #212, opened 2022-07-07
UI contacts

popkinj:
## Enhancements
- Wallet information
- Table of contacts
- Button for adding new contact
- Page component layout

## Adjustments
- Moved the `add contact` component into a modal

**Recommending a Squash Merge.**


popkinj:
Definitely need to squash merge this sucker.

services/grip/frontend/.env.test
```diff
@@ -0,0 +1 @@
+VITE_TRACTION_ENDPOINT=https://traction-api-test.apps.silver.devops.gov.bc.ca:5100
```
loneil, line 1: We'll have to adjust how the environmental settings for Traction URLs get into the app at deployment time when we figure out our hosting and Helm/OCP plans
https://apps.nrs.gov.bc.ca/int/jira/browse/TRACTION-314

services/grip/frontend/.env.development
```diff
@@ -0,0 +1 @@
+TRACTION_ENDPOINT=https://traction-api-dev.apps.silver.devops.gov.bc.ca:5100
```
loneil, line 1: There's a few of these "No newline at end of file" warnings

services/grip/frontend/src/components/Contacts.vue
```diff
@@ -0,0 +1,105 @@
+<script setup lang="ts">
+import { inject, ref } from "vue";
+import Fieldset from "primevue/fieldset";
+import DataTable from "primevue/datatable";
+import Column from "primevue/column";
+import ProgressSpinner from "primevue/progressspinner";
+import axios from "axios";
+import Button from "primevue/button";
+import Dialog from "primevue/dialog";
+import CreateContact from "./CreateContact.vue";
```
loneil, line 10: Not important but would be nice to group the primevue imports in order together

services/grip/frontend/.env.test
popkinj: Yeah. I'm starting to think we should either configure NGinx to proxy the API or maybe spin up a little Node server to serve the assets and proxy.

services/grip/frontend/src/components/Contacts.vue
popkinj: Agreed... that would be much cleaner. I'll check to see if Rollup can handle that.

PR #213, opened 2022-07-07
Feature/v1 holder offers

usingtechnology:
Add database model for Holder Credentials/Offers.
Include protocol listener to create when offered, update status when accepted, and update the revocation notifier to read from new table.

Added a new "base" class to include tags and external_reference_id, updated a few classes that should inherit from it.


loneil (approved):
Looks good to me.

PR #214, opened 2022-07-08
Feature/tenant UI make issuer

Jsyro:
Button to mke issuer
if innkeeper has not allowed it, user get's toast message with 409 error

No webhook or handlers, so hard refresh is needed to see updates. 

After set as issuer, show disabled button explaining. could be displayed in a better way in the future.


Jsyro:
image

loneil (approved):
Pulled locally, tried fail and success case. Looks good (works in `npm run dev` mode fine for this one as well)
Minor spacing comment

PR #215, opened 2022-07-08
API and services for Holder Credentials

usingtechnology:


loneil (commented):
Looks good. On the deployed PR

- Created issuer/holder tenants
- Set up issuer and cred
- connected the 2
- Issued cred from issuer to holder connection
- Checked status of cred then accepted it on holder side
- Another cred and rejected

PR #216, opened 2022-07-13
Issuer status not update on reject (Openshift only)

usingtechnology:
Troubleshooting... no key error on 'state'...


loneil (approved):
Tried issuer>holder accepted and rejected and both work and return the correct Traction status/state now

image

image

Saw note about workaroud vs aca-py version so can discuss later. But just confirming this does work in current state.

PR #217, opened 2022-07-14
Feature/tenant scope queries

Jsyro:
Remove SA native queries, use BaseModel or new TenantScopedModel.tenant_select() to automatically add the tenant_id from the starlette_context to the query. 

Tasks and Webhook callbacks will also need to be tenant_scoped, those procedures have a tenant_id, and can be integrated into this solution later, for now, the existing code will include a 'fallback_tenant_id' that will be used if there isn't a starlette_context to read. 

Will make safe_update, and safe_delete after feedback

And will write tests for this behaviour now too.


services/traction/api/db/models/base.py
```diff
@@ -3,16 +3,20 @@
...
 from sqlmodel import Field, SQLModel
 from sqlalchemy import Column, func, text, select, desc, String
+from sqlalchemy.sql.expression import Select
 from sqlalchemy.dialects.postgresql import UUID, TIMESTAMP, VARCHAR, ARRAY
 from sqlmodel.ext.asyncio.session import AsyncSession
 
 from api.db.session import async_session
 from api.endpoints.models.v1.errors import NotFoundError
 
+logger = logging.getLogger(__name__)
```
usingtechnology, line 18: can we make the logger a class var, so we can determine which class/table is making the log?

services/traction/api/db/models/base.py
```diff
@@ -23,6 +27,14 @@ class Config(pydantic.BaseModel.Config):
 class BaseModel(SQLModel, BaseSchema):
     __mapper_args__ = {"eager_defaults": True}
 
+    @classmethod
+    def tenant_select(cls, fallback_tenant_id: UUID = None) -> Select:
```
usingtechnology, line 31: Do not like this... I know you are protecting from developer error, but we should be catching that in the review - that the model class is not using the TenantScopedModel. If a developer has an IDE that uses intellisense or the like, this will appear as a valid method to call, and it shouldn't be.

services/traction/api/db/models/base.py
```diff
@@ -65,6 +77,29 @@ async def update_by_id(cls, item_id: UUID, values: dict):
                 raise e
 
 
+class TenantScopedModel(BaseModel):
+    tenant_id: uuid.UUID = Field(foreign_key="tenant.id", index=True)
+
+    @classmethod
+    def tenant_select(cls, fallback_tenant_id: UUID = None) -> Select:
+        from api.endpoints.dependencies.tenant_security import get_from_context
```
usingtechnology, line 85: will have to think about this, seems weird to have endpoints dependency in the db. but probably just a refactor when we have other contexts... should leave a TODO here that this is a dependency smell

services/traction/api/db/models/base.py
```diff
@@ -65,6 +77,29 @@ async def update_by_id(cls, item_id: UUID, values: dict):
...
+        tenant_context_id = get_from_context("TENANT_ID")
+        # load from starlette context
+        if tenant_context_id:
+            result = select(cls).where(cls.tenant_id == tenant_context_id)
+        # manually passed in fallback for now
+        elif fallback_tenant_id:
+            result = select(cls).where(cls.tenant_id == fallback_tenant_id)
+        # couldn't load tenant context for some reason
+        else:
+            logger.error("not a http context, no tenant_id, EXECUTING AN UNSAFE QUERY")
```
usingtechnology, line 97: maybe make this a warning. and if the logger belongs to a class, we should be able to track it down

services/traction/api/db/models/base.py
```diff
@@ -146,7 +181,11 @@ async def list_by_item_id(
...
         """
 
-        q = select(cls).where(cls.item_id == item_id).order_by(desc(cls.created_at))
+        q = (
+            cls.tenant_select()
+            .where(cls.item_id == item_id)
+            .order_by(desc(cls.created_at))
+        )
         q_result = await db.execute(q)
         db_items = q_result.scalars()
```
usingtechnology, line 190: might as well fix the scalars().all()

loneil (commented):
I've done regression tests on innkeeper/contact/issue/hold stuff and can't see any issues. Will wait for Sherman to give a review as well of code changes.

usingtechnology (changes requested):
`TenantConfiguration` and `TenantAutoResponseLog` are used by tenant and have `tenant_id` column creation, so those can use `TenantScopedModel`. (`TenantPermissions` is an innkeeper table so should not).

All the db model methods that take tenant_id as a parameter, that parameter should be Optional - that parameter should only be passed in if it is not in a context (like in a task for now).

And for the services, wouldn't we want `tenant_id`/`wallet_id` to be optional there too? if we are plucking and passing those values around everywhere, then we aren't gaining anything except for a query/where helper method (which is valuable on its own for sure). 

If we aren't going there (making use of the context), then I think we could just make this PR the `TenantScopedModel` class and use `tenant_select` in classes that use it. We can come back later and address using the context(s) and not passing tenant_id/wallet_id around.

services/traction/api/db/models/base.py
Jsyro: I just want someone who copy pastes an example to not get a 'method not found' and instead the code tells them what their issue is. 

@loneil @popkinj, thoughts?

services/traction/api/db/models/base.py
Jsyro: It was just to avoid a circular import, if we moved the tenant_security stuff somewhere else we could probably fix this. 

I'll give it a shot

Jsyro:
> All the db model methods that take tenant_id as a parameter, that parameter should be Optional - that parameter should only be passed in if it is not in a context (like in a task for now).
> 
> And for the services, wouldn't we want `tenant_id`/`wallet_id` to be optional there too? if we are plucking and passing those values around everywhere, then we aren't gaining anything except for a query/where helper method (which is valuable on its own for sure).
> 
> If we aren't going there (making use of the context), then I think we could just make this PR the `TenantScopedModel` class and use `tenant_select` in classes that use it. We can come back later and address using the context(s) and not passing tenant_id/wallet_id around.


Totally agree, we shouldn't even allow tenant_id to be passed in by the developer, the context should handle it all. But I didn't want to deal with that given we wanted to make this change fast and iterate on it later. 

The model GET methods require the tenant_id, because the context may not be set if it's a task or a webhook handler. so those go hand in hand. 

Wallet_id could totally be handled in this way as well. 

I want the code to be used, and can be improved on. I think that's better for adoption or improvement than keep it as a special case until it's perfect.

services/traction/api/db/models/base.py
Jsyro: fixed

services/traction/api/db/models/base.py
usingtechnology: it may not work without abstract classes.  so ok to leave for now I guess... perhaps we can just be careful with the log statements to include the table name.

services/traction/bdd-tests/features/steps/security.py
```diff
@@ -0,0 +1,26 @@
+import requests
+import json
+
+from behave import *
+from starlette import status
+
+
+@then(
+    '"{tenant_1}" cannot NOT get "{tenant_2}"s contact to "{tenant_2_contact_alias}" by id'
```
loneil, line 9: Should this read "can NOT get" rather than the double negative? It's testing that tenant 1 isn't allowed to see tenant2's contacts right?

services/traction/api/endpoints/dependencies/tenant_context.py
```diff
@@ -0,0 +1,13 @@
+from starlette_context import context
+from starlette.exceptions import HTTPException
+from starlette import status
+
+
+def get_from_context(name: str):
+    result = context.get(name)
+    if not result:
```
usingtechnology, line 8: i don't think throwing an HTTP exception when querying the database level is appropriate.

services/traction/api/endpoints/routes/v1/tenant/context_route.py
```diff
@@ -0,0 +1,30 @@
+# from typing import Callable
```
usingtechnology, line 1: does this work? if so, why aren't we using it?

services/traction/api/endpoints/routes/credentials.py
```diff
@@ -24,7 +24,7 @@
 
 
 from api.endpoints.dependencies.db import get_db
-from api.endpoints.dependencies.tenant_security import get_from_context
+from api.endpoints.dependencies.tenant_context import get_from_context
```
usingtechnology, line 27: i don't understand why we move a function to a new file like this?

services/traction/bdd-tests/features/steps/security.py
```diff
@@ -0,0 +1,26 @@
...
+)
+def step_impl(context, tenant_1: str, tenant_2: str, tenant_2_contact_alias: str):
+
+    response1 = requests.get(
```
usingtechnology, line 13: these requests are in `steps/v1_api.py` trying to reduce duplication.

services/traction/bdd-tests/features/steps/security.py
```diff
@@ -0,0 +1,26 @@
...
+        context.config.userdata.get("traction_host") + "/tenant/v1/contacts",
+        params={"alias": tenant_2_contact_alias},
+        headers=context.config.userdata[tenant_2]["auth_headers"],
+    )
+    assert response1.status_code == status.HTTP_200_OK, response1.__dict__
+    tenant_2_contact = json.loads(response1.content)["items"][0]
+
+    response2 = requests.get(
```
usingtechnology, line 21: i agree that both list/search and get need to be tested, but probably better if list (tenant 1) and list (tenant 2) are compared and get (tenant 1) and get (tenant 2) are compared.

usingtechnology (commented):
still lots to discuss.

services/traction/api/db/models/v1/connection_invitation.py
```diff
@@ -87,7 +86,6 @@ class ConnectionInvitation(BaseModel, table=True):
     async def get_by_id(
         cls: "ConnectionInvitation",
         db: AsyncSession,
-        tenant_id: UUID,
```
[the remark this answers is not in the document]
Jsyro: That is the goal, but 'setting the context' should not need to be done by every individual method, it should be handled globally by middleware or the event bus. 

Working on this though.

services/traction/bdd-tests/features/steps/security.py
Jsyro: I don't see the difference, if the lists do no have 1 item in them, something went wrong. and the get by id should always go 404. 

Are you thinking there might be some asymmetry somewhere. Like Tenant2 could read Tenant1's data, but not visa versa?

Don't want to write unnecessary tests. and if we do want to be more granular, it should not be in the BDD test suite,

usingtechnology: those calls, list/search and get are completely different stacks. so you should be comparing the same stack. you are not confirming tenant 2's list function cannot return the value. likewise, you are not confirming that tenant 1's get works. you could do list 1 by alias like you have, get 1 using the id from list 1. then list 2 should fail (either nothing found with that alias, or that item has a different id than from list 1), and get 2 should fail as you have above.

services/traction/api/endpoints/routes/credentials.py
Jsyro: because tenant_security imports modesl.base.py now, so it's a circular import if they are in the same file

usingtechnology: ah, ok. regardless that tenant_context.get_from_context doesn't really apply at the db level (HTTP exception saying not authenticated isn't appropriate - or rather misleading)

services/traction/bdd-tests/features/steps/security.py
Jsyro: There are two tests, one is that the list endpoints only return their own contacts. 

The other grabs an ID from another tenants (arrange), and queries the get_by_id endpoint with a tenant that does not own that object (act), and asserts it's a 404. 

i have to get the id somehow, i could save it locally from the setup, but this seemed easier to understand, and easier to write.

usingtechnology: a more complete test actually calls both tenants with the same calls. in the above, you do not. it's not a lot of work to add 1 contact per tenant. call list for both tenants, their results should have 1 and their ids should be different. now you have 2 contacts from 2 tenants. then tenant 1 calls get with their contact id and tenant 2's contact id (404), likewise tenant 2 calls with their contact id and tenant 1's contact id (404). i wouldn't list by alias, i'd assert those after the fact to make sure each tenant's resulting contact is the one they are supposed to have.

PR #218, opened 2022-07-14
Holder Presentation DB models and basic state tracking

usingtechnology:

Going to go with a small, functional PR. This is the basis for completing Holder functionality in v1, just adding the db model and the protocol listener that will create a record and update the state from AcaPy events.

No API endpoints added, this is the next PR.


services/traction/api/db/models/v1/holder.py
```diff
@@ -286,3 +286,214 @@ async def get_by_revocation_ids(
...
+        db: AsyncSession,
+        tenant_id: uuid.UUID,
+        thread_id: str,
+    ) -> List["HolderPresentation"]:
+        """List by Thread ID.
+
+        Find and return list of Holder Presentation records for Thread ID.
+
+          tenant_id: Traction ID of tenant making the call
+          thread_id: AcaPy Thread ID of Issuer Presentation
```
loneil, line 485: Is this supposed to read "Holder Presentation"? Or is Issuer right for this one...

usingtechnology: nope, copy/paste...

usingtechnology: fixed.

loneil (approved):
Looks good

PR #219, opened 2022-07-15
Fix Tenant UI qr code bug and add structure for AIP compatibility page.

loneil:
1. Fix an issue causing the QR code to not show in `npm run dev` mode (ref needs to set value). Not sure why it worked in `build` mode but still does.
2. Add the structure for the AIP compatibility page. See https://github.com/bcgov/traction/pull/219/files#diff-9852fb872af0af579b0f5fe635e7d6eea4638580a09dbcd4cfe4f911216b505a. Just basic structure now, then we can edit it in GH without needing to do PRs as we go on

### DIFF
--- a/README.md
+++ b/README.md
@@ -56,6 +56,10 @@ Overview of the traction architecture:
 
 [Architecture](./docs/ARCHITECTURE.md)
 
+Aries Interop Profile status:
+
+[Traction AIP](./docs/AIP.md)
+
 Source code for Traction API:
 
 [Traction](./services/traction/README.md)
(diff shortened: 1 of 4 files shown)


popkinj (approved):
Can't really comment on the AIP stuff. But the QRCode component looks good.

README.md
```diff
@@ -56,6 +56,10 @@ Overview of the traction architecture:
 
 [Architecture](./docs/ARCHITECTURE.md)
```
usingtechnology, line 57: 👍

docs/AIP.md
```diff
@@ -0,0 +1,122 @@
+# traction aries interop profile
```
usingtechnology, line 1: 👍 👍

usingtechnology (approved):

PR #220, opened 2022-07-16
Feature/v1 holder presentation api

usingtechnology:
Add API/services for Holder Presentations.

Includes BDD tests.


loneil (approved):
Code changes look good to me. Endpoint structure makes sense I think. Some other regression testing seems fine. Didn't pull locally for BDD but if they're working on yours. 

I think we'll want to prioritize 7.4 update next sprint so we don't have do deal with in-between stuff for testing and such.

PR #221, opened 2022-07-18
List schemas 328

popkinj:
# New component that lists available schemas.
- Temporarily pointing to localhost API
- There is a test schema file commented out that I'll need for the sprint review (I still can't add schemas locally).
- There are a couple form elements (button and text input) that will get activated in future tickets.


services/grip/frontend/public/test_schemas.json
```diff
@@ -0,0 +1,107 @@
+{
```
loneil, line 1: Do we want to commit this under public? 
If it's not needed for the app might not want it. Could be useful test data under some test/fixtures directory maybe?

popkinj: This file is just temporary. I need it for the demo of the sprint review then I'll blow it away.
Do you want me to add it to the .gitignore?

loneil: Yeah I'd leave it locally if it's just for a specific demo session on one local machine. But it's useful to have test data for unit tests and other things so could consider moving it to a test folder location too.

popkinj: Do you want test/fixtures to be at the top level @loneil ? Or within Grip?

PR #225, opened 2022-07-20
Feature/tenant select method

Jsyro:
Centralise code that adds tenant_id filter to all queries. 

Will revisit smarter tenant security later.


usingtechnology (commented):
That base class (TenantScopedModel) with tenant id is a great idea.
Providing an example of testing for tenant restriction is also a great idea.

As the code stands, it is more confusing rather than less. 
There is a custom select method which can only be enforced by review. The fact that we grab and pass the tenant_id in the round pass it to services then vanishes is confusing. 

Having an HTTPException with an authorization message is misleading and db level shouldn't be throwing an HTTPException. I don't think we have any contextmanager/yielded code, but if there is and it ends up using the db in its context/yield, then an HTTPException bad practice (ie. totally borks the thread).

Providing the example test is great, but as we discussed this example doesn't provide a robust test of the restrictions.

PR #237, opened 2022-07-28
Tenant UI structure

loneil:
Serve the tenant UI from a Node API that provides config, and proxies Traction calls through to the configured Traction environment.

Just added a new "tenant-ui" services folder and will remove the "grip" one later.

Lots of files here but you can probably ignore most everything in the frontend components and views folders since those are the existing controls without changes and will be changed around a lot for the new UI layout.

The main layout has a number of debug/development things at the top, will leave those for the moment since they're helpful to understand the interactions but will remove next with UI setup.

Prettier/eslint stuff is not behaving (and seems like kind of a nightmare with Volar/vue/typescript so that will need a bit of attention later)

See tenant-ui/README.md for instructions.


usingtechnology (approved):
looks like a great start to me

Jsyro (approved):
This looks good!

PR #241, opened 2022-08-02
Add some more TenantUI components and refresh contacts on creation

loneil:
Added schemas table, some credentials tables, refresh contacts when you create an invitation.


usingtechnology (approved):
LGTM.
Good pattern to establish for the table refresh. My VUE is rusty, took me a bit to follow it, but I finally got it.

loneil:
> LGTM. Good pattern to establish for the table refresh. My VUE is rusty, took me a bit to follow it, but I finally got it.

@usingtechnology  Yeah emit will work well for child components in a simple way. If there's complex views with a bunch of siblings can expand to the store state.

PR #244, opened 2022-08-04
set helm charts to deploy.

usingtechnology:
update to first valid image


Jsyro (approved):
Looks Good!

PR #245, opened 2022-08-05
Feature/showcase removal

usingtechnology:
This PR removes Showcase code and CICD and code in Traction API that was only used by Showcase.
A note has been added to the main README file showing how to get a working copy of Showcase.


Jsyro (approved):
My favourite kind of pull request. as long as BDD tests pass i'm good with this!

PR #247, opened 2022-08-09
View My Received Presentations

Jsyro:
These are things you have proven about your contacts. aka, presentation-exchanges where you are the 'verifier'.

I see the existing pattern is to include the contact details in the object. I'm not sure i like that pattern when it could be dealt with on the FE. 

If including the contact details (and not just the contact) is to be standard, please let me know. 


usingtechnology (approved):
Looks good to me.

The result set should match the others though (list issued credentials, list holder presentations, list holder credentials). The API is not for `tenant-ui`, it's for other applications. We should not expect that other apps have cached the contacts, or expect they want to do N+1 calls to fill a table. Returning the contact id, the contact's external reference id (in case they need to lookup a better name in their system) and the alias is not returning a full contact. But it allows callers to reasonably fill a table with the human readable contact (alias).

PR #248, opened 2022-08-12
Send UX environment variables for configuration

usingtechnology:

We need to support deploying "custom" instances, so need to get build/deploy time configuration passed into the Frontend. Starting simple with the application title and a few others that will help with troubleshooting (image details, traction url) . We will expand with things such as color when we sort out scss and themes.


popkinj (approved):
Looks good. :)

PR #250, opened 2022-08-17
Fix innkeeper check-in, add tenant list to innkeeper API

usingtechnology:

Quick touch-up to fix a bug in the innkeeper `check-in` API.
Added in a function (list tenants) that was not brought forward from v0 API.


popkinj (approved):
Yeah... Doesn't mean much to me. But, I don't see any red flags. :)

PR #251, opened 2022-08-18
Use Pinia for storage.

usingtechnology:
Put API calls in store.
Update configuration from backend to include proxy path.


popkinj:
Oops. Accidentally took this out of draft.
I'm OK with using Pinia in the short term. Vue3 has moved away from third party stores and towards the [Composition API](https://vuejs.org/guide/extras/composition-api-faq.html). This is the officially supported format and comes with lots of optimizations. It also provides all the functionality that Pinia and Vuex do.

I'm also seeing lots of potential merge conflicts. Moving forward we need to coordinate who is working on which files. As it stands I'm going to have to roll back some logic, such as the token storage, to get back to the vue3 way of things.

popkinj:
I'm also noticing a lot of unnecessary synchronous behavior. Freezing the app for asynchronous events is not a very good user experience.

usingtechnology:
> I'm also noticing a lot of unnecessary synchronous behavior. Freezing the app for asynchronous events is not a very good user experience.

yeah, good catch. just coming from my brain in a world of `awaits` everywhere. i have removed most of them here, need one on `create-invitation` as noted in the code. we can just keep an eye out on times we need an await and handle it then/there.

usingtechnology:
> Oops. Accidentally took this out of draft. I'm OK with using Pinia in the short term. Vue3 has moved away from third party stores and towards the [Composition API](https://vuejs.org/guide/extras/composition-api-faq.html). This is the officially supported format and comes with lots of optimizations. It also provides all the functionality that Pinia and Vuex do.
> 
> I'm also seeing lots of potential merge conflicts. Moving forward we need to coordinate who is working on which files. As it stands I'm going to have to roll back some logic, such as the token storage, to get back to the vue3 way of things.

[pinia](https://vuejs.org/guide/scaling-up/state-management.html#pinia) is "owned" by vuejs, so not really 3rd party, so reasonable for this project at this time. and I wanted the things that populate the stored data (the API) as part of the store, wasn't really seeing how best to achieve that with reactivity - but I'm a n00b for Vue 3... 

as far as conflicts... has to happen sometime if we are figuring out common patterns, better now than when there are 4 of us.  having common patterns established could help reduce conflicts 🤷

popkinj:
> 

Yeah... Promises and callbacks are good for not tying up the event loop.

popkinj:
> 

Agreed about the conflict thing. Better now then later.
... And I took a look at pinia and it is using the Composition API under the hood. So I think we're good there.

popkinj (approved):
Can't see anything wrong. Let'r fly.

PR #254, opened 2022-08-23
Invalid Token - traction api return 401.

usingtechnology:
Tenant UI can differentiate between server error (500) and token error (401)


loneil (approved):
👍

PR #255, opened 2022-08-24
Accept invitation screen and validation library

loneil:
Simple accept invitation form with validation
image

On submit/success it just does a info toast, could consider navigating back to the contacts list after as well... will wait for UX design and feedback from users/etc.

On the component did not use the onAction store handler, just try/catch the awaited pinia action, this pattern can be used for simple case components to keep the flow in one place. The subscrption (onAction) style will be useful for more complex callbacks or integrating multiple components.

Validation syntax is really verbose and I think there can be improvements there as we learn the PrimeVue/Vuelidate libraries and interactions better.

### DIFF
--- a/services/tenant-ui/frontend/package.json
+++ b/services/tenant-ui/frontend/package.json
@@ -10,6 +10,8 @@
     "lint": "eslint --ext .ts,vue --ignore-path .gitignore ."
   },
   "dependencies": {
+    "@vuelidate/core": "^2.0.0-alpha.44",
+    "@vuelidate/validators": "^2.0.0-alpha.31",
     "axios": "^0.27.2",
     "pinia": "^2.0.19",
     "primeflex": "^3.2.1",
--- a/services/tenant-ui/frontend/src/components/contacts/AcceptInvite.vue
+++ b/services/tenant-ui/frontend/src/components/contacts/AcceptInvite.vue
@@ -1,0 +1,61 @@
+<template>
+    <h3 class="mt-0 mb-5">Accept Invitation</h3>
+    <div class="form-demo">
+        <form @submit.prevent="handleSubmit(!v$.$invalid)">
+            <div class="field">
+                <div class="p-float-label">
+                    <Textarea id="inviteUrl" v-model="v$.inviteUrl.$model"
+                        :class="{ 'p-invalid': v$.inviteUrl.$invalid && submitted }" :autoResize="true" rows="1"
+                        cols="100" />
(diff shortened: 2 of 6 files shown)


services/tenant-ui/frontend/src/store/contactsStore.ts
```diff
@@ -59,7 +59,44 @@ export const useContactsStore = defineStore('contacts', () => {
...
+      .then((res) => {
+        console.log(res);
+        // don't grab the item, there are other parts of the response data we need (invitation, invitation url)
+        accepted_data = res.data;
+        console.log(accepted_data);
+      })
+      .then(() => {
+        // do we want to automatically reload? or have the caller of this to load?
+        console.log('invitation accepted. the store calls load automatically, but do we want this done "manually"?');
+        listContacts();
```
loneil, line 80: Reloading after as with the create contact part in the store itself. I think this'll be the pattern to stick with but can see as we go.

services/tenant-ui/frontend/package.json
```diff
@@ -10,6 +10,8 @@
     "lint": "eslint --ext .ts,vue --ignore-path .gitignore ."
   },
   "dependencies": {
+    "@vuelidate/core": "^2.0.0-alpha.44",
```
usingtechnology, line 13: ALPHA!!! yikes... 😉

services/tenant-ui/frontend/src/store/contactsStore.ts
usingtechnology: totally... we will see what patterns emerge.  contacts is probably the safest to "auto" refresh as it will be in pick lists everywhere.

services/tenant-ui/frontend/src/components/contacts/AcceptInvite.vue
```diff
@@ -0,0 +1,98 @@
...
+// Form submission
+const submitted = ref(false);
+const handleSubmit = async (isFormValid: boolean) => {
+    submitted.value = true;
+
+    if (!isFormValid) {
+        return;
+    }
+
+    try {
```
usingtechnology, line 88: 👍 this pattern is nice and tight and succinct. easy to know what is going on.

usingtechnology (approved):
good stuff!

services/tenant-ui/frontend/package.json
loneil: Hey the guy is on vacation!
https://github.com/vuelidate/vuelidate/issues/1078#issuecomment-1171959960

Jsyro (approved):
Looks clean to me!

PR #256, opened 2022-08-25
TRACTION-371 - form to create cred def (template)

usingtechnology:
will work for either created or copied schema. limiting to a single cred template for now.
hack in place to list cred templates with schema templates, add new ticket to put in API.
form looks bad... styling and making layout universal is another ticket, this is purely functional.

### DIFF
--- a/services/tenant-ui/frontend/src/components/issuance/Schemas.vue
+++ b/services/tenant-ui/frontend/src/components/issuance/Schemas.vue
@@ -23,12 +23,22 @@
           </span>
         </div>
       </template>
+      <Column :sortable="false" header="Actions">
+      </Column>
       <Column field="name" header="Schema" filter-field="name" />
       <Column field="version" header="Version" />
       <Column field="status" header="Status" />
       <Column field="state" header="State" />
       <Column field="attributes" header="Attributes" />
       <Column field="schema_id" header="ID" />
+      <Column field="credential_templates" header="Credential Template">
+        <template #body="{data}">
+          <CreateCredentialTemplate :schema-template-id="data.schema_template_id" v-if="data.credential_templates.length == 0"/>
+          <div v-else>
+            {{`${data.credential_templates[0].name}:${data.credential_templates[0].tag}`}}  
+          </div>
+        </template>
+      </Column>
     </DataTable>
     <div class="row buttons">
       <Button
@@ -75,6 +85,7 @@ import Dialog from "primevue/dialog";
 // Custom components
 import CreateSchema from "./CreateSchema.vue";
 import CopySchema from "./CopySchema.vue";
+import CreateCredentialTemplate from "./credentialtemplate/CreateCredentialTemplate.vue";
 
 import { useToast } from "vue-toastification";
 import { useGovernanceStore } from "../../store";
--- a/services/tenant-ui/frontend/src/components/issuance/credentialtemplate/CreateCredentialTemplate.vue
+++ b/services/tenant-ui/frontend/src/components/issuance/credentialtemplate/CreateCredentialTemplate.vue
@@ -1,0 +1,32 @@
+<template>
+    <Button
+        icon="pi pi-id-card" 
+        class="p-button-text"        
+        @click="displayForm = true"
+        v-tooltip.top="'Create Credential Template'"
+    />
+    <Dialog header="Create Credential Template" v-model:visible="displayForm">
+        <CreateCredentialTemplateForm 
+            @form-close="displayForm = false;"
+            :schema-template-id="schemaTemplateId"
+        />
+    </Dialog>
+</template>
+<script setup lang="ts">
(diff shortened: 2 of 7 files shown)


services/tenant-ui/frontend/src/components/issuance/Schemas.vue
```diff
@@ -23,12 +23,22 @@
           </span>
         </div>
       </template>
+      <Column :sortable="false" header="Actions">
+      </Column>
```
loneil, line 27: Can just be <Column :sortable="false" header="Actions" />

services/tenant-ui/frontend/src/components/issuance/Schemas.vue
```diff
@@ -23,12 +23,22 @@
...
       <Column field="name" header="Schema" filter-field="name" />
       <Column field="version" header="Version" />
       <Column field="status" header="Status" />
       <Column field="state" header="State" />
       <Column field="attributes" header="Attributes" />
       <Column field="schema_id" header="ID" />
+      <Column field="credential_templates" header="Credential Template">
+        <template #body="{data}">
+          <CreateCredentialTemplate :schema-template-id="data.schema_template_id" v-if="data.credential_templates.length == 0"/>
```
loneil, line 36: I think can leave off the ` == 0` for the boolean expression (`v-if="data.credential_templates.length")

services/tenant-ui/frontend/src/components/issuance/credentialtemplate/CreateCredentialTemplate.vue
```diff
@@ -0,0 +1,32 @@
...
+        v-tooltip.top="'Create Credential Template'"
+    />
+    <Dialog header="Create Credential Template" v-model:visible="displayForm">
+        <CreateCredentialTemplateForm 
+            @form-close="displayForm = false;"
+            :schema-template-id="schemaTemplateId"
+        />
+    </Dialog>
+</template>
+<script setup lang="ts">
```
loneil, line 15: Annoying style quibble but I'd love to keep a space between template, script, and style blocks

services/tenant-ui/frontend/src/components/issuance/credentialtemplate/CreateCredentialTemplateForm.vue
```diff
@@ -0,0 +1,160 @@
+<template>
+    <h3 class="mt-0 mb-5">Create Credential Template</h3>
+    <div class="form-demo">
```
loneil, line 3: Doesn't look like the 'form-demo' class is applying anything anywhere?
This could be a good place (or on the form element itself) to apply a global 'tenant-ui-form' class that we apply global style to all sub elements to make form spacing what we want, have some thoughts about that for the near future)

services/tenant-ui/frontend/src/store/governanceStore.ts
```diff
@@ -8,20 +8,91 @@ export const useGovernanceStore = defineStore('governance', () => {
...
+
+  const schemaTemplateDropdown = computed(() => {
+    //transform list of schemas into label / value pair
+    let result: any[] = [];
+    if (schemaTemplates.value != null) {
+      result = schemaTemplates.value
+        .map((item: any) => schemaLabelValue(item));
+    }
+    return result;
+  });
```
loneil, line 41: Awesome, almost the exact same code I have locally for the contact dropdown getter (might be able to have some more general utility that maps the dropdown but that could be more work than it's worth since the field names will be specific)

services/tenant-ui/frontend/src/store/governanceStore.ts
```diff
@@ -8,20 +8,91 @@ export const useGovernanceStore = defineStore('governance', () => {
...
+      // find any credential templates...
+      const templates = ref();
+      await fetchList('/tenant/v1/governance/credential_templates/', templates, error, loading, {schema_template_id: item.schema_template_id});
+      return {
+        ...item,
+        credential_templates: templates
+      };
+    }));
+    return schemaTemplates.value;
+  }
```
loneil, line 91: Yeah hack for expedience so we can iterate on this but yeah I dont think we'd want to do the cred def fetch in the map since that's looping that API call for each schema then. Could do the 2 lists _first_ and then map them together by ID, but like you said this might be a job for the API

loneil (commented):
This is great stuff, leaving comments now, will just poke some testing after the standup before approving

loneil:
@usingtechnology could make this form field show disabled/greyed if it's a primevue input property (or maybe wait til global form style)

image

loneil (approved):
Awesome stuff, approving, comments are mostly app-wide format/style stuff we need to address. Though we might want to think about tracking that looping API call fix since if we forget about it a user with tons of schemas (if possible) could be making a lot of api calls for creds.

PR #257, opened 2022-08-26
Delete schema 339

popkinj:
- Unstyled button for deleting schemas
Screenshot from 2022-08-26 12-16-26


services/tenant-ui/config/local.json
```diff
@@ -4,6 +4,6 @@
     "version": "latest"
```
usingtechnology, line 4: can we just ignore this file please? everyone seems to want something different...

services/tenant-ui/frontend/src/store/governanceStore.ts
```diff
@@ -8,7 +8,7 @@ export const useGovernanceStore = defineStore('governance', () => {
   const schemaTemplates: any = ref(null);
```
usingtechnology, line 8: i guess we need to figure out a standard linter/formatter... this will get annoying...

usingtechnology (approved):
the basics are done.
are we doing another pr to prompt before delete (establish the pattern for the next delete X function)?

services/tenant-ui/config/local.json
loneil: I thought I added it to gitignore? It should be gitignored and we shouldn't be checking it in

services/tenant-ui/frontend/src/store/governanceStore.ts
loneil: Yeah mine doesn't do this format here when I auto-format... (alt-shift-f) though maybe it _should_ be based on the prettier rules? Not sure but we definitely need a quick look at getting that stuff standardized

services/tenant-ui/config/local.json
popkinj: Yeah, I was thinking the prompt for "are you sure" before deletion was a separate ticket.

services/tenant-ui/frontend/src/store/governanceStore.ts
popkinj: Agreed. If we can have all our editors read linter rules from the repo .linter files (instead of local env) that would make our lives a lot easier.

loneil (approved):

PR #259, opened 2022-08-29
Offer a credential

loneil:
Offer a credential functionality. Should be working end-to-end except the part where you fill in the cred right now is still raw JSON. Would like to get this merged in for simplicity and so we can be trying out the issuance flow and continue to try at the pretty multi-field entry next.

Includes auto-completes for populating credential and contact.

OfferCredentialForm is where most of the work is done, it makes use of multiple stores. Even though this is a issuer component a lot of the work is done by the governance store since the existing methods to handle all the schema/cred stuff are handled there, which is probably ok but can look over store structure again when we have a lot of components in and interacting.

The credential value dance between the edit/display/POST (multiple variables, a little confusing) can possibly be improved but will be altered significantly when the dynamic text boxes go in.

Did package updates in this PR as well.

image

Have some work in progress to replace this part below but need to get this in now first.
image

image


services/tenant-ui/frontend/src/components/issuance/offerCredential/OfferCredentialForm.vue
```diff
@@ -0,0 +1,228 @@
...
+      <!-- Main Form -->
+      <div v-if="!showEditCredValues">
+        <!-- Credential -->
+        <div class="field">
+          <label for="selectedCred" :class="{ 'p-error': v$.selectedCred.$invalid && submitted }">Credential
+            Name
+            <ProgressSpinner v-if="credsLoading" />
+          </label>
+
+          <AutoComplete id="selectedCred" v-model="v$.selectedCred.$model" :disabled="credsLoading"
```
usingtechnology, line 13: slick!

services/tenant-ui/frontend/src/store/issuerStore.ts
```diff
@@ -18,7 +21,40 @@ export const useIssuerStore = defineStore('issuer', () => {
     return fetchList('/tenant/v1/issuer/credentials/', credentials, error, loading);
   }
 
-  return { credentials, selectedCredential, loading, error, listCredentials };
+  async function offerCredential(payload: any = {}) {
+    console.log('> issuerStore.createSchemaTemplate');
```
usingtechnology, line 25: wrong method name... not a blocker though

usingtechnology (approved):
looks great!
if you have time, you wanna throw in the table refreshes? would be super useful.
not going to block without it since we will be doing more on the form itself ASAP

PR #266, opened 2022-09-02
Format dates in tables

loneil:
Was going to use a filter as in Vue2 apps to format data like this in templates but that's not a Vue3 thing:
https://v3-migration.vuejs.org/breaking-changes/filters.html#_3-x-update

So using a helper method just imported into the components as needed. Thought about adding a global one to the Vue globalProperties but read some best practices stating global Vue things are getting a lot more discouraged (import as needed into composition API)

Added date-fns library to deal with date parsing and formatting.

image

### DIFF
--- a/services/tenant-ui/frontend/src/components/contacts/Contacts.vue
+++ b/services/tenant-ui/frontend/src/components/contacts/Contacts.vue
@@ -65,7 +65,7 @@ import Dialog from 'primevue/dialog';
 import ProgressSpinner from 'primevue/progressspinner';
 import InputText from 'primevue/inputtext';
 // State
-import { useContactsStore } from '../../store';
+import { useContactsStore } from '@/store';
 import { storeToRefs } from 'pinia';
 // Other imports
 import { useToast } from 'vue-toastification';
--- a/services/tenant-ui/frontend/src/components/contacts/Contacts.vue
+++ b/services/tenant-ui/frontend/src/components/contacts/Contacts.vue
@@ -29,7 +29,11 @@
       <Column field="role" header="Role" />
       <Column field="state" header="State" />
       <Column field="status" header="Status" />
-      <Column field="created_at" header="Created at" />
+      <Column field="created_at" header="Created at">
+        <template #body="{ data }">
+          {{ formatDateLong(data.created_at) }}
+        </template>
+      </Column>
       <Column field="contact_id" header="ID" />
     </DataTable>
   </div>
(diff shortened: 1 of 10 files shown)


services/tenant-ui/frontend/src/components/contacts/Contacts.vue
```diff
@@ -53,23 +57,21 @@
...
-import { useToast } from 'vue-toastification';
+// State
 import { useContactsStore } from '../../store';
 import { storeToRefs } from 'pinia';
-
+// Other imports
+import { useToast } from 'vue-toastification';
 // Other components
 import CreateContact from './CreateContact.vue';
+import { formatDateLong } from '@/helpers';
```
popkinj, line 74: Like to see there are some Vue tools for this... Instead of bringing in something larger like moment.js.

services/tenant-ui/frontend/src/helpers/index.ts
```diff
@@ -0,0 +1,33 @@
+import { format, parseJSON } from 'date-fns'
```
popkinj, line 1: Oh.... haha. I see you used this thing for date formatting. Still... Must be much smaller then other tools.

popkinj (approved):
Let there be readable dates!! ☺

usingtechnology (approved):
nice work, finally adding a little polish to this stuff!

(i added the format to the presentation details component).

services/tenant-ui/frontend/src/helpers/index.ts
usingtechnology: > Oh.... haha. I see you used this thing for date formatting. Still... Must be much smaller then other tools.

there is [Intl.DateTimeFormat](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/DateTimeFormat) (native) that maybe we can use? maybe pin that for later, could take a while to figure out.

usingtechnology: but that leads to a better question... should the traction api be returning a string or a numeric value for timestamps.  probably... both? at least with numeric it's much easier for callers to format themselves.

loneil: Yeah Moment is deprecated now (https://momentjs.com/docs/#/-project-status/) and most of the Googling about what to do instead suggests date-fns.

Wrapping it in that helper can let us swap it out for anything else in the future too

PR #269, opened 2022-09-02
Feature/presentation details claims order

Jsyro:
Preserve claims order by using the request order instead of the presentation order. Add clean getters to separate the 4 cases.


loneil (approved):
Yeah if this gets Kyle the order/layout desired for the demo then 👍 can look at a lot more when more UX decisions are finalized

PR #271, opened 2022-09-02
attempt to have all modals operate the same way.

usingtechnology:
not overly concerned with style, have not added validation


services/tenant-ui/frontend/src/components/issuance/copySchema/CopySchemaForm.vue
```diff
@@ -0,0 +1,93 @@
...
+    toast.info('Schema copied');
+  } catch (error) {
+    toast.error(`Failure: ${error}`);
+  } finally {
+    submitted.value = false;
+  }
+};
+</script>
+
+<style>
```
loneil, line 87: Scope this, seems to have messed up some other forms site-wide (offer credential dropdowns)

loneil:
The reason the schema modals aren't closing as you'd want when you submit the form the way the Issue Credential one is because the issue cred one actually contains the components inside the table area that is redrawn on loading, so that clears everything out when a successful reload of the div happens.

You could move the buttons inside the v-else div a couple lines up. Or better, moving them into the table header would match the design mockups.

So the closing on submit behaviour (in the place it works) is really just a quirk of the redrawing. The other way could be to do the emit pattern on the successful submit so the parent component of the form (the button/dialog pair) knows the form is done, or in the parent handle the form action resolution.

I tried moving it into the table header and that worked so I can commit if you want (didn't want to in case you have local changes that would conflict).

Also something wierd is on in this branch where the form elements even on the offer credential form are not sized properly now... can look into that.

usingtechnology:
closing... revisited this in [277](https://github.com/bcgov/traction/pull/277).

PR #272, opened 2022-09-06
Style front page and theme

loneil:

Get front page laid out the way the mockups are
https://xd.adobe.com/view/20ae96da-b8b6-4dca-8d17-da1c3403613f-75fd/screen/3eff8eb3-d749-429c-868b-f5b68f3d8323/
It's responsive to screen size (image disappears at lower breakpoints). Some finicky aspect ratio stuff with the image at lower sizes but can try and tweak around with that later if it's noticeable.
Added validation to login form

I **left off** the things on the login page mockups that we do not have guidance text (policy etc) or functionality for.

Adjusted primary color so the buttons and table highlights and stuff are the correct color from the mockups. Did this by copying the theme being used and adjusting sass (also added sass preprocessing) as described here https://stackoverflow.com/a/66395579/324672
This means we can centralize color/spacing/sizing changes for primevue elements in just the one place.

Linted/build/docker built and tested existing functionality after sass/theme change and things seem fine.

image


usingtechnology (approved):
looks so much better, the sass is a step in the right direction too.

popkinj (approved):
Looks great!

PR #273, opened 2022-09-06
Presentation Details demo tweak

Jsyro:
Some contextual information on what the user is looking at
image

Add an accordion below data for 'identifiers' which include cred_defs and schema_ids. 
image

### DIFF
--- a/services/tenant-ui/frontend/src/components/verifier/PresentationDetails.vue
+++ b/services/tenant-ui/frontend/src/components/verifier/PresentationDetails.vue
@@ -16,6 +16,21 @@
         <li>Contact Alias: {{ presentation.contact.alias }}</li>
         <hr />
       </div>
+
+      <!-- VERIFIED Meaning-->
+      <div v-if="props.show_information && presentation.status == 'verified'">
+        <span class="pi pi-check"></span
+        ><span
+          >Credential is held by
+          <strong>{{ presentation.contact.alias }}</strong></span
+        ><br />
+        <span class="pi pi-check"></span><span>Credential is valid</span><br />
+        <span class="pi pi-check"></span><span>Credential is tamper-free </span
+        ><br />
+        <span class="pi pi-check"></span
+        ><span>All attribute restrictions were statisfied</span><br />
+      </div>
+      <hr />
       <!-- PRESENTATION RECEIVED-->
       <!-- requested_attributes using 'names' list -> revealed attribute groups -->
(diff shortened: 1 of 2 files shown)


services/tenant-ui/frontend/src/components/verifier/PresentationDetails.vue
```diff
@@ -16,6 +16,21 @@
...
+        ><span
+          >Credential is held by
+          <strong>{{ presentation.contact.alias }}</strong></span
+        ><br />
+        <span class="pi pi-check"></span><span>Credential is valid</span><br />
+        <span class="pi pi-check"></span><span>Credential is tamper-free </span
+        ><br />
+        <span class="pi pi-check"></span
+        ><span>All attribute restrictions were statisfied</span><br />
+      </div>
```
loneil, line 32: Kind of a wierd format with these span braces here. This our format settings causing it? I've seen the old vue extension do this and can clean it up by putting some manual newlines between sibling elements.

services/tenant-ui/frontend/src/components/verifier/PresentationDetails.vue
```diff
@@ -53,6 +68,21 @@
         <!-- requested_attribute using 'name' string w/o restrictions -> revealed self-attested values -->
         <!-- requested_predicates -> unrevealed attributes -->
       </div>
+      <hr />
+      <!-- Identifiers -->
+      <Accordion>
+        <AccordionTab
+          v-bind:header="`Identifier_${index + 1}`"
+          v-for="(item, index) in props.presentation.acapy.presentation_exchange
```
loneil, line 76: Would a lint complain about no ':key' on this one?

Jsyro: yes it did, fixed.

services/tenant-ui/frontend/src/components/verifier/PresentationDetails.vue
Jsyro: the formatter is doing that. is there a better way to do this than span tags and  <br/> tags?

loneil: No I think the tags are fine, it's just the line length when the tags are next to each other without a space can cause a wierd line break I think

loneil (approved):

PR #275, opened 2022-09-06
Feature/disable inaccessible buttons

Jsyro:
Disable buttons that require the tenant to be an issuer. 

image

image

Button behaviour/look is unchanged when tenant is an issuer.


loneil (commented):
A few lint warnings
image

loneil (changes requested):
My test tenant (TEST env) is an issuer and I've issued but the button is disabling on this branch

image

image

loneil (approved):
Works for me now

Jsyro:
messed up the rebase, redoing in new branch.

PR #277, opened 2022-09-07
Feature/modal cleanup revisted

usingtechnology:
Have all modals operate similarly, use a parent component with a button and dialog and then the form component. 
Handle open and closing in the parent component.

Had to move some styles that were accidentally global into the main scss, we will revisit when we do a style sweep and clean up.

Removed progress spinners in favour of using the table `:loading`, this allows us to refresh the table data without reloading child components of the spinner (ie the dialogs!).

Removed types from schema attribute creation.

Did not work on styling up the actual forms very much, leaving that for the big clean up, just wanted the dialogs to work consistently and setup some patterns.


services/tenant-ui/frontend/src/assets/_theme.scss
```diff
@@ -5694,3 +6465,24 @@ $tenant-ui-shadow: lighten( $tenant-ui-secondary-color, 30% );
...
+.field > label,
+.field > small.p-error {
+  display: block !important;
+}
+
+.field > label > .p-progress-spinner {
+  height: 1em;
+  width: 1em;
+}
+/* <<<<<<<<<<< */
```
loneil, line 6488: I'd keep these out of the theme file (there's the style.css and we can do some reconfigure of that to be scss as well)

services/tenant-ui/frontend/src/components/contacts/Contacts.vue
```diff
@@ -1,82 +1,62 @@
...
+        {{ formatDateLong(data.created_at) }}
+      </template>
+    </Column>
+    <Column field="contact_id" header="ID" />
+  </DataTable>
 </template>
 
 <script setup lang="ts">
 // Vue
-import { ref, onMounted } from 'vue';
```
popkinj, line 59: Good catch here.

popkinj (approved):
Looks good. More organized this way.

PR #278, opened 2022-09-07
fix the innkeeper tenants listing,

usingtechnology:
Bugs on matching when parameter was None.
Need the option to include or exclude the issuer status, public did status and issuer flag from filtering. 
This call is very weird (right now), as we don't store that data, so we need to populate it in code and the filter afterward.


loneil (approved):
Tried it out on swagger with a couple tenants and filters

PR #279, opened 2022-09-07
Feature/disable issuer buttons if not issuer

Jsyro:
try 2. Offer Credential, Create Schema, and Create Credential Definition are disabled if `issuer` boolean is set to false. 

image


loneil (approved):
Works for me

PR #281, opened 2022-09-09
Clean up menu items, page titles, etc

loneil:
Feedback from Kyle on here https://apps.nrs.gov.bc.ca/int/jira/browse/TRACTION-416

Also, add vite configuration to expose our theme variables to all components.
(https://github.com/bcgov/traction/pull/281/files#diff-dc567cb65ef7671c99f55d0daf37e71642121520e369c93cf8feaa5af989fe13)

Menu item for schemas etc is "Configuration", not sure about that one.... we have Tenant configuration, though we're calling that "Settings" in Jamie's branch work going on so maybe ok. Can alter again some time.

Note changes in screenshot below
image


services/tenant-ui/frontend/src/assets/variables.scss
```diff
@@ -0,0 +1,7 @@
+// Theme variables
```
usingtechnology, line 1: smart, adding this file. nice to see things moving to a more maintainable direction.

services/tenant-ui/frontend/src/router/index.ts
```diff
@@ -47,55 +40,51 @@ const routes = [
   },
 
   {
-    path: '/issuance/',
+    path: '/configuration/',
```
usingtechnology, line 43: do we want to move configuration down the menu? it's important but seems odd placed between contacts/issuer/verifier/holder... not important, just something that looked "off" to me.

usingtechnology (approved):
great stuff, good progress and much cleaner and consistent.

sorry that I jammed in some changes to your PR... (just got ALL the tables to behave the same way...)

services/tenant-ui/frontend/src/router/index.ts
loneil: Yeah, the ticket didn't have any placement for it, but I'd agree, moving

popkinj (approved):
Was worried this would conflict with my branch... But all looks good. :)

PR #282, opened 2022-09-09
Request issuar 410

popkinj:
- User profile menu
Screenshot from 2022-09-07 15-52-25
- Button for requesting access to Issuance permission.
- Logout button works in a fashion that will be replaced when our router is fixed.


popkinj (commented):
Oh thanks... I forgot to run the lint thing.

loneil (approved):
Tested out with and without permissions and functionality works, can see where design will go from here.

PR #284, opened 2022-09-09
Feature/presentation details 2

Jsyro:
Table layout, no predicate support yet. Relevant UX mockup. https://xd.adobe.com/view/20ae96da-b8b6-4dca-8d17-da1c3403613f-75fd/screen/ad2e839c-1b83-4500-88f5-ff5c5e4fda01/

Show requested attributes (groups, single, and self-attested) in one table. Checkmark to indicate the data was verified against something, info icon brings up modal with more details on the restrictions. 

image

restrictions infoModal
image


usingtechnology (approved):
Conditionally approve this. will let others review, we shouldn't be leaving @ts-expect-error comments but needed them so I could build and review.

Jsyro:
> Conditionally approve this. will let others review, we shouldn't be leaving @ts-expect-error comments but needed them so I could build and review.

Thanks for catching the linting, forgot again, we should look at making that a GH action.

Unsure about the `@ts-expect-error` usage, doing some research now.

.vscode/settings.json
```diff
@@ -5,4 +5,16 @@
...
+      "previewLimit": 50,
+      "server": "localhost",
+      "port": 5432,
+      "driver": "PostgreSQL",
+      "name": "docker-traction",
+      "username": "postgres",
+      "password": "postgresPass",
+      "database": "traction"
+    }
+  ],
```
loneil, line 19: Don't think this was to be comitted?

services/tenant-ui/frontend/src/components/verifier/VerifiedPresentationData.vue
```diff
@@ -0,0 +1,155 @@
...
+  return Object.fromEntries(
+    Object.entries(
+      props.presentation.acapy.presentation_exchange.presentation_request
+        .requested_attributes
+    ).filter(([key, ra]) => {
+      // @ts-expect-error types have not been defined for this object
+      return 'name' in ra && !('restrictions' in ra);
+    })
+  );
+};
```
loneil, line 149: There's probably a bunch of utility-function-ing that can be done here to reduce it, but I'd be inclined to leave it until we done have as much in-flight or proposed changes still to Presentation UX.

.vscode/settings.json
Jsyro: It's in the settings file and it's still valid for other users. If we don't want this in the repo then we should un-version the file.

I can remove it if desired.

services/tenant-ui/frontend/src/components/verifier/VerifiedPresentationData.vue
Jsyro: Yea, i know the JS here is rudimentary, but i'm also confident that any developer can understand this code. Let me know if you want me to make any changes here now.

.vscode/settings.json
loneil: I'd clear it if it's dev machine specific tooling

loneil (approved):

.vscode/settings.json
Jsyro: added to ignore, but this change was already in an earlier commit.

PR #285, opened 2022-09-11
Update base readme to include tenant ui.

usingtechnology:
Update Use Case to use tenant ui, update videos to match current state.


docs/USE-CASE.md
```diff
@@ -1,95 +1,126 @@
-# Running the traction demo
+# Traction Demo
 
-## Running traction using swagger
+The Traction demo utilizes the Innkeeper API (tenant manager), the Tenant API and Tenant UI.  Tenant UI is currently under development and does not implement all the features of the Tenant API. We will be hopping around between Swagger and the Tenant UI to show:
```
loneil, line 3: Maybe note that all the stuff being done with the Tenant UI is doable with raw API calls?
I think we want to drive home that this interface is optional and you can integrate through another app with calls. Maybe it's clear enough elsewhere?

usingtechnology: will do - good call.

PR #286, opened 2022-09-12
Fielded credential attribute entry

loneil:
When entering cred values dynamically display input fields.
Allow a JSON editor as well to drop in the raw **attribute** JSON block (the same thing the control had before). You can toggle back and forth and the 2 modes will update themselves. If the JSON entered has invalid syntax (no nice schema validation or anything, this is power user mode) then toast a warning and don't update anything.

Also add delete contact.

image

image

image

image

image

### DIFF
--- a/services/tenant-ui/frontend/src/components/contacts/Contacts.vue
+++ b/services/tenant-ui/frontend/src/components/contacts/Contacts.vue
@@ -27,6 +27,16 @@
     </template>
     <template #empty> No records found. </template>
     <template #loading> Loading data. Please wait... </template>
+    <Column :sortable="false" header="Actions">
+      <template #body="{ data }">
+        <Button
+          title="Delete Contact"
+          icon="pi pi-times-circle"
+          class="p-button-rounded p-button-icon-only p-button-text"
+          @click="deleteContact($event, data)"
+        />
+      </template>
+    </Column>
     <Column :sortable="true" field="alias" header="Name" />
     <Column field="role" header="Role" />
     <Column field="status" header="Status" />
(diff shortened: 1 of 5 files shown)


usingtechnology (approved):
that's great!
we can extract the raw json to a component on the next entry screen we do... think that will be useful to crank screens out without waiting on design.

Jsyro:
I'm unsure about the behaviour of a JSON syntax error causing the user to loose the form, and not have any changes applied. If the use case is that the user is going to be copy-pasting something in, then it's not that 'destructive' but if they manually edited it, then lost their edit, that could be annoying. Maybe check the syntax before switching views, but then a novice user might get stuck in the JSON view and I don't like that either. 

Just something to consider, not sure there is a right or wrong answer.

popkinj (approved):
Nice work @loneil .
What do you guys think about incorporating a json editor/linter like [JSON Editor](https://jsoneditoronline.org/#left=local.vawovu)? May be easier on the newbies. Something to think about for later maybe.

usingtechnology:
My $0.02 is that we document the above concerns (JSON error handling, JSON Editor) and implement in the next component - we need to crank out place holders forms for view and edit and we can tackle it there. I think we should be popping in a basic open row with View Raw JSON like @Jsyro has in his PR, and also an entry form that is just the JSON editor so we can get more raw functionality into Tenant UI and not hop around to Swagger so much.

loneil:
Yeah the JSON edit on this is just a power user copy/paste attributes thing in (per Kyle's request). So just letting you know if there's a syntax error instead of a generic error was all I wanted. Going back and forth between fielded and JSON and then making an mistake in JSON isn't really an edge case there's a good ux handle on at this point. For example you can go into JSON and copy paste attributes that aren't actually in the schema and we're not doing schema compare validation here or anything at this point.

Jsyro (approved):
This looks good, makes me think about the future of handling deleted contacts.

PR #289, opened 2022-09-13
Issue cred loading button on submit, image copyright

loneil:
Forgot to add the spinner when submitting a cred offer, so it doesn't look like the form is stalled while the post call goes.
Also need copyright image on the cover image. Can't add to images since it can be off the screen based on aspect ratio so overlay. Added structure for this as a user configurable option.

### DIFF
--- a/charts/tenant-ui/templates/tenant_ui_configmap.yaml
+++ b/charts/tenant-ui/templates/tenant_ui_configmap.yaml
@@ -12,3 +12,4 @@ data:
   UX_SIDEBAR_TITLE: {{ .Values.tenant_ui.ux.sidebarTitle | quote }}
   UX_COPYRIGHT: {{ .Values.tenant_ui.ux.copyright | quote }}
   UX_OWNER: {{ .Values.tenant_ui.ux.owner | quote }}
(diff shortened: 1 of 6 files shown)


usingtechnology:
Very weird, I can lint and build without issue, but running frontend `npm run dev` I had an issue with:

```
The following dependencies are imported but could not be resolved:

  vue-json-pretty/types/utils (imported by /Users/jason/github/bcgov/traction/services/tenant-ui/frontend/src/components/issuance/offerCredential/EnterCredentialValues.vue?id=0)
```

removing the unused import "fixed" it...

```
import { emitError } from 'vue-json-pretty/types/utils';
```
not sure if you hit anything like that.  what's even weirder is putting that line back in hot-loaded just fine. 🤷 

Should probably remove that line; maybe others will encounter the same weirdness.

loneil:
@usingtechnology thanks for finding, didn't get any error on my end so might chalk it up to weird Vite stuff that happens sometimes. Regardless it's unused so should come out yeah 👍

Jsyro (approved):
LGTM!!

PR #291, opened 2022-09-15
Enable history mode server-side

loneil:
Fall back to serve the vue app if a non-defined route is hit. This enables Vue Router history mode https://router.vuejs.org/guide/essentials/history-mode.html#html5-mode

Added a very light commonly used middleware lib for this on the backend https://www.npmjs.com/package/connect-history-api-fallback
Also recommended on Vue Router docs (https://router.vuejs.org/guide/essentials/history-mode.html#express-with-node-js)

So what happens now is you can refresh on the front end and it will go back to the route rather than 404ing. But the token in state is lost so you get bounced through login again.
 This is a starter for the final behaviour we'll want but want to get the PR in for this functionality. More improvements can be:
- Keeping your token locally so refreshing doesn't kill your login session
- Nav gaurds for the vue routes to go to login if token is invalidated
- Proper front end 404 pages

Tested this with

- Local dev mode
- Local start mode
- Docker container
- OCP route

### DIFF
--- a/services/tenant-ui/frontend/src/store/configStore.ts
+++ b/services/tenant-ui/frontend/src/store/configStore.ts
@@ -33,7 +33,7 @@ export const useConfigStore = defineStore('config', () => {
     error.value = null;
     loading.value = true;
     await axios
-      .get('config')
+      .get('/config')
       .then((res) => {
         console.log(res);
         config.value = res.data;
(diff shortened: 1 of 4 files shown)


usingtechnology (approved):
👍 smart first step

PR #292, opened 2022-09-15
Generic Row Expand showing Raw JSON

usingtechnology:

Generic component we can put in tables now that will show the` View Raw Content` JSON expander. The component has one named slot that can be filled in with specific fancier UI layout of the item.


services/tenant-ui/frontend/src/components/common/RowExpandData.vue
```diff
@@ -0,0 +1,33 @@
...
+</template>
+
+<script setup lang="ts">
+import { computed, PropType, ref } from 'vue';
+import Accordion from 'primevue/accordion';
+import AccordionTab from 'primevue/accordiontab';
+import VueJsonPretty from 'vue-json-pretty';
+import 'vue-json-pretty/lib/styles.css';
+import { JSONDataType } from 'vue-json-pretty/types/utils';
+import { processSlotOutlet } from '@vue/compiler-core';
```
loneil, line 22: Looks unused

services/tenant-ui/frontend/src/components/common/RowExpandData.vue
```diff
@@ -0,0 +1,33 @@
...
+    <Accordion>
+      <AccordionTab header="View Raw Content">
+        <vue-json-pretty :data="data" />
+      </AccordionTab>
+    </Accordion>
+  </div>
+</template>
+
+<script setup lang="ts">
+import { computed, PropType, ref } from 'vue';
```
loneil, line 16: computed, ref unused

services/tenant-ui/frontend/src/components/contacts/Contacts.vue
```diff
@@ -105,6 +126,11 @@ const doDelete = (schema: any) => {
       toast.error(`Failure: ${err}`);
     });
 };
+
+const expandedRows = ref([]);
+const onRowExpand = async (event: any) => {
+  await getFullContact(event.data.contact_id);
+};
```
loneil, line 133: Can we pass the composable down into the expand component? So some of the generic work is being done there instead of in each table component.
Like the component knows it's opened (onMount maybe...) and then it invokes the supplied method to get data?

loneil: Then you don't have to pass in loading and data, the sub component would have that stuff encapsulated in it's own scope? I haven't thought it through entirely so could be some stopper for that but a thought.

usingtechnology: was working on that. it needs to be in a sub-component anyway, or else every row is binding to the same item (one referenced in Contacts.vue.  So there will be a wrapper component that will fetch it's own data then make use of expand row...

services/tenant-ui/frontend/src/components/layout/AppLayout.vue
```diff
@@ -41,7 +41,7 @@ body {
   display: flex;
   flex-direction: row;
   align-items: flex-start;
-  height: 100vh;
+  min-height: 100vh;
```
loneil, line 44: I just pushed these to get the sidebar color to stretch down when the page expands (was constraining it to only fill the page height when it should fill AT LEAST the page height)

usingtechnology: 👍

services/tenant-ui/frontend/src/components/contacts/Contacts.vue
usingtechnology: now have examples for contact, schema and verifier presentation. those are self contained (ie. the table component doesn't do anything around fetching data...)

services/tenant-ui/frontend/src/components/verifier/PresentationRowExpandData.vue
```diff
@@ -0,0 +1,73 @@
...
+  showInformation: {
+    type: Boolean as PropType<boolean>,
+    required: false,
+    default: false,
+  },
+});
+
+const { loading, item, fetchItemWithAcapy } = useGetVerifierPresentation();
+
+fetchItemWithAcapy(props.row.verifier_presentation_id);
```
Jsyro, line 70: This also removes the behaviour of storing these detailed loads in the store? 

Is this the new pattern going forward, i thought we had discussed extending the refresh of this table to clear that detail location in the store.

Jsyro (commented):
Code looks great, one behavioural difference I wanted to clarify.

loneil (approved):
So far I like, there might be some more efficiency around passing the call in and not needing to handle data and loading by the parent but I can't figure it out right now.

Supportive of keeping it simple and lazy loading data as needed for this model. Doesn't risk the issue where a table refresh isn't displaying refreshed expended data then.

usingtechnology:
> So far I like, there might be some more efficiency around passing the call in and not needing to handle data and loading by the parent but I can't figure it out right now.
> 
> Supportive of keeping it simple and lazy loading data as needed for this model. Doesn't risk the issue where a table refresh isn't displaying refreshed expended data then.

yeah, was also going to say something similar.. feels like the composable can get refactored a lot.

services/tenant-ui/frontend/src/components/verifier/PresentationRowExpandData.vue
usingtechnology: correct, just simplified things for now and we can come back to cache when we need it. once the mechanism is figured out, should be able to slot it in very easily.

usingtechnology:
adding JIRA Ticket for refactoring the composable... TRACTION-441

PR #297, opened 2022-09-27
Api payload 437

popkinj:
Component to send a custom JSON payload to the API. Accepts URL and a JSON template as Props.

Screenshot from 2022-09-27 09-42-18
Screenshot from 2022-09-27 09-42-28
Screenshot from 2022-09-27 09-43-07
Screenshot from 2022-09-27 09-44-36

### DIFF
--- a/services/tenant-ui/frontend/package-lock.json
+++ b/services/tenant-ui/frontend/package-lock.json
@@ -17,6 +17,7 @@
         "primeicons": "^5.0.0",
         "primevue": "^3.17.0",
         "qrcode.vue": "^3.3.3",
+        "vanilla-jsoneditor": "^0.7.5",
         "vue": "^3.2.39",
         "vue-json-pretty": "^2.2.2",
         "vue-router": "^4.1.5",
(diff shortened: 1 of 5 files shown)


loneil:
Works for me as a start (might need some additional configuration as we add in a few places so the button can be specific labeling if we need). But we should take it off the Contacts screen, as that already has the actions it needs for invitation management at the moment. Could just throw in the Presentation Templates page with this one as a starter
https://traction-api-test.apps.silver.devops.gov.bc.ca/tenant/docs#/verifier/create_presentation_request_template_v1_verifier_presentation_templates__post

popkinj:
> Works for me as a start (might need some additional configuration as we add in a few places so the button can be specific labeling if we need). But we should take it off the Contacts screen, as that already has the actions it needs for invitation management at the moment. Could just throw in the Presentation Templates page with this one as a starter https://traction-api-test.apps.silver.devops.gov.bc.ca/tenant/docs#/verifier/create_presentation_request_template_v1_verifier_presentation_templates__post

Sure. That may be a better example seeing as it has a larger payload. I'll move it over after the demo.

Jsyro (commented):
Not sure about the name `SuperYou` but otherwise looks great.

PR #298, opened 2022-09-27
Profile and Settings pages

loneil:
Add the profile and settings pages. Lifted the "Make Issuer" button over as it already was, new ticket coming to add UI improvements for Issuer flag from Guru's mockups
https://xd.adobe.com/view/20ae96da-b8b6-4dca-8d17-da1c3403613f-75fd/screen/84591ecc-ddc8-4485-92ef-136e6c6c6f57

Store Messages and Store Issuer Credentials buttons will only work if the innkeeper has allowed that for your tenant but there's no way of the tenant having that info, so probably some future improvements there.

Also library updates


popkinj (approved):
May have a bit of a merge conflict with the npm lock file... But looks good otherwise.

PR #300, opened 2022-10-03
Feature/accept rej cred offer

Jsyro:
When in 'offer received' give user buttons to accept or reject cred offer. 

Currently there is no UI to inspect the details of the offered credentials, that is out of scope for this ticket. 

image

after clicking accept

image

after clicking delete 

image

after confirmation

image

### DIFF
--- a/services/tenant-ui/frontend/src/components/holder/Credentials.vue
+++ b/services/tenant-ui/frontend/src/components/holder/Credentials.vue
@@ -59,13 +59,21 @@
           <Button
             v-tooltip.top="'Delete Credential'"
             label="Delete"
-            class="p-button-danger"
+            class="p-button-secondary"
             icon="pi pi-times-circle"
             @click="deleteCredential($event, data)"
           />
         </div>
       </template>
     </Column>
+    <Column :sortable="true" field="alias" header="Name" />
+    <Column field="status" header="Status" />
+    <Column field="created_at" header="Created at">
+      <template #body="{ data }">
+        {{ formatDateLong(data.created_at) }}
+      </template>
+    </Column>
+    <Column field="contact.alias" header="Contact Name" />
   </DataTable>
 </template>
 
--- a/services/tenant-ui/frontend/src/components/holder/Credentials.vue
+++ b/services/tenant-ui/frontend/src/components/holder/Credentials.vue
@@ -104,12 +104,12 @@ const { loading, credentials, selectedCredential } = storeToRefs(
 
 const acceptOffer = (event: any, data: any) => {
   holderStore.acceptCredentialOffer(data.holder_credential_id).then(() => {
-    loadTable();
+    toast.success(`Credential successfully added to your wallet`);
   });
 };
 const rejectOffer = (event: any, data: any) => {
   holderStore.rejectCredentialOffer(data.holder_credential_id).then(() => {
-    loadTable();
+    toast.success(`Credential offer rejected`);
   });
 };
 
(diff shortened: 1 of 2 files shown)


loneil:
Like you said no specific UX yet so can alter later but could consider moving the Actions column to the front since that's where it is on other screens/mockups
image

Jsyro:
> Like you said no specific UX yet so can alter later but could consider moving the Actions column to the front since that's where it is on other screens/mockups. Could go with icon buttons instead of the block text ones to be consistent too. image

moved to left, and made reject different than delete. 

image

services/tenant-ui/frontend/src/components/holder/Credentials.vue
```diff
@@ -46,17 +74,44 @@ import { useToast } from 'vue-toastification';
...
 
 const holderStore = useHolderStore();
 // use the loading state from the store to disable the button...
 const { loading, credentials, selectedCredential } = storeToRefs(
   useHolderStore()
 );
 
+const acceptOffer = (event: any, data: any) => {
+  holderStore.acceptCredentialOffer(data.holder_credential_id).then(() => {
+    loadTable();
```
usingtechnology, line 92: this isn't the same pattern as used elsewhere for having the table refreshed on add/delete/etc. for good or ill, they should be the same for now.

services/tenant-ui/frontend/src/store/holderStore.ts
```diff
@@ -35,6 +36,36 @@ export const useHolderStore = defineStore('holder', () => {
     );
   }
 
+  const tenantApi = useTenantApi();
+
+  async function acceptCredentialOffer(cred_id: string) {
+    await tenantApi
```
usingtechnology, line 42: Same with these functions - don't match up with similar existing calls... 
Did we decide to change the pattern up? Cool if we have, but I don't see other stores having been updated. 🤷

services/tenant-ui/frontend/src/components/holder/Credentials.vue
loneil: I would support calling the refresh function after these as well, the table should do it's thing after you accept/delete and then you'd see the updated entry.

Jsyro: It appears the existing pattern is for the store operation to update the related affected objects. 

So the table get's update because the store updates itself, rather than leaving it to the component to manage the refresh. Just something i missed, however that store code itself could use some simplification\.

usingtechnology: yeah, we don't really know what we want to do with the store and data anyway. I'm not super convinced we need to store anything but the token/creds for the tenant.  the store design allows for a bunch of different ways of interacting with the async stuff, like the same function can be used as we are using it, and also support subscribing to 'onAction'. this pattern has emerged, so we should strip out the extra stuff to support subscriptions...

loneil (approved):
Functionally works for me, would still consider whether we should change the button style to be in line with the mockups https://xd.adobe.com/view/20ae96da-b8b6-4dca-8d17-da1c3403613f-75fd/screen/11e80784-cb21-4421-a2e5-181d6c5095f2

but also fine with waiting until Guru can do a pass over it.

Jsyro:
> Functionally works for me, would still consider whether we should change the button style to be in line with the mockups https://xd.adobe.com/view/20ae96da-b8b6-4dca-8d17-da1c3403613f-75fd/screen/11e80784-cb21-4421-a2e5-181d6c5095f2
> 
> but also fine with waiting until Guru can do a pass over it.

image

loneil (approved):

PR #301, opened 2022-10-03
Add row data expand for holder and issuer credentials

usingtechnology:

Added in row expand for tables that were missing them.
Will refactor in a subsequent ticket... had something going just passing in the URL to fetch but getting schemas is actually two API calls :(


loneil:
Seems to be opening all of them when I click one of them (Issued Creds at least)

image

usingtechnology:
> Seems to be opening all of them when I click one of them (Issued Creds at least)

well, that seems like a bug!

loneil (approved):

PR #302, opened 2022-10-04
Return cred templates with schema template (optional)

usingtechnology:

Address issues in tenant UI where we need the list of credential templates for each schema template. We were calling list schemas and list credential templates and merging the lists.

Update the traction API to optionally return credential templates per schema template.
Update tenant UI to make the single call and ask the API to return credential templates.

This is a pre-cursor PR so we can refactor tenant UI row expand logic - reduce a bunch of duplicated code. Schema/Credential Template was the only operation that required logic in the UI to build the result set. Now that the API call returns the complete result set, we can do a refactor.


popkinj (approved):
:+1:

PR #303, opened 2022-10-04
Innkeeper UI Structure

loneil:
Innkeeper UI/entrance to the app. UX (color/text/images/etc) to come so just reviewing this for the basic flow.

Navigate to normal Tenant UI path and add /innkeeper to access

**Routing/Views**
Reconfigure routing to track all **tenant** side pages as children of / and **innkeeper** ones as children of /innkeeper. Add views and components for the innkeeper side.

I copied the AppLayout component to reuse for Innkeeper, this is probably a bit of duplication, thought it could be kept as 1 file and then parameterize the components of that through Router Named Views, but thought it was prudent to allow for a different layout if they diverge with subsequent UX changes and tweaks, so we'll have a separate Innkeeper layout component. I think I'll want to extract the duplicated layout css though in a future PR when I do the styling when UX is done. Same for duplicated Sidebar.

Change the main app entry to a router view that uses the default parent View of / or /innkeeper to determine which layout and login to use, AppLayout (or InnkeeperAppLayout) have the containing router view for the children routes.

**State**
Add an InnkeeperAPI store as we do for the tenant API (as comment in tenant api says can possibly look later at making that a service rather than a store, but working great for now so maybe not). Add an innkeeper token store to contain the auth.

The api store has a lot of duplication from the Tenant one with the REST Verb methods, can possibly extract that into a general thing but will leave for refactoring in case we need a different pattern after reviews.

At this point the auth, in state, is still nullified on any reload or URL nav. There's no router way of 'crossing between' the tenant and innkeeper worlds, nor should there be. So if you go from one to another your token is lost from local memory, which is good for security. Soon I'd like to **keep** your token locally for preserving login sessions so you don't have to reenter PW on a refresh etc. If that's implemented we'll want to add something in the Tenant Auth that nullifies any stored innkeeper token, and vice versa.

### DIFF
--- a/.github/workflows/on_pr_opened.yaml
+++ b/.github/workflows/on_pr_opened.yaml
@@ -140,6 +140,7 @@ jobs:
             | | Innkeeper Swagger | https://pr-${{ github.event.number }}-traction-api-dev.apps.silver.devops.gov.bc.ca/innkeeper/docs |
             | | Tenant Swagger | https://pr-${{ github.event.number }}-traction-api-dev.apps.silver.devops.gov.bc.ca/tenant/docs |
             | | Tenant UI | https://pr-${{ github.event.number }}-tenant-ui-dev.apps.silver.devops.gov.bc.ca |
+            | | Innkeeper UI | https://pr-${{ github.event.number }}-tenant-ui-dev.apps.silver.devops.gov.bc.ca/innkeeper |
             | Aca-Py | | |
             | | Admin Swagger | https://pr-${{ github.event.number }}-traction-acapy-admin-dev.apps.silver.devops.gov.bc.ca/api/doc |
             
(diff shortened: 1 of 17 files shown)


Jsyro (approved):
Looks great to me, any chance you can quickly change the header color for the innkeeper so it is more visually distinguishable. Not a huge deal, just would be nice if you have two tabs open to not get confused.

.github/workflows/on_pr_opened.yaml
```diff
@@ -140,6 +140,7 @@ jobs:
             | | Innkeeper Swagger | https://pr-${{ github.event.number }}-traction-api-dev.apps.silver.devops.gov.bc.ca/innkeeper/docs |
             | | Tenant Swagger | https://pr-${{ github.event.number }}-traction-api-dev.apps.silver.devops.gov.bc.ca/tenant/docs |
             | | Tenant UI | https://pr-${{ github.event.number }}-tenant-ui-dev.apps.silver.devops.gov.bc.ca |
+            | | Innkeeper UI | https://pr-${{ github.event.number }}-tenant-ui-dev.apps.silver.devops.gov.bc.ca/innkeeper |
```
usingtechnology, line 143: 👍

usingtechnology (approved):
Good stuff.
Yeah, lot of duplication that we can (attempt to) clean up later. No need to worry about it right now.

PR #304, opened 2022-10-05
Reduce duplicated code for Row Expand components

usingtechnology:

Uses API paths coded into the table, do we want to make constants for all API paths? or wait until the store/service pattern for our API calls is hammered out?


services/tenant-ui/frontend/src/components/common/RowExpandData.vue
```diff
@@ -2,32 +2,45 @@
...
+    default() {
+      return {};
+    },
   },
 });
+
+const { loading, item, fetchItem } = useGetItem(props.url);
+
+// ok, let's load up the row with acapy data...
+fetchItem(props.id, props.params);
```
loneil, line 45: I'd normally be doing a fetch on an onMounted lifecycle event (or a watched prop or something). I might have just missed it in Vue3 but is this a typical pattern? Loading just as part of the setup script? It seems to be working.

Jsyro (approved):
Love that added/deleted lines ratio!!! Haven't tested, but looks good.

loneil:
Path constants could be useful even now. Whether you want to do all the replacement as part of this PR or leave it is ok.

services/tenant-ui/frontend/src/components/common/RowExpandData.vue
usingtechnology: Got this pattern when I was looking into composables and loading data asynchronously. The setup script is async, so it'll render when ready.

loneil: https://vuejs.org/api/sfc-script-setup.html#top-level-await

loneil (approved):

PR #305, opened 2022-10-06
Innkeeper styling

loneil:
Get the innkeeper theming in. There's some more to do/figure out but we will iterate on it as we get more UX designs going (Guru has a mockup set started that I'm working off).

Re-aranged the scss to separate concerns/spit files for easier maintenance as well.

image

image


usingtechnology (approved):
looks great, nice work getting scss rolling.

PR #306, opened 2022-10-06
Prevent creation of credential template if schema template not active.

usingtechnology:
Prevent use of credential template for issuance if not active.


loneil (approved):
Pulled locally and looking good to me

PR #307, opened 2022-10-07
Tenant management in Innkeeper UI

loneil:
In the Innkeeper UI, add the list of tenants and the ability to check in a new tenant.

Some of the is still waiting on UX so wording/layout subject to change in a future PR.

**Tenants list**
On the main section of innkeeper load the tenants with row expander for full details (will get feedback on columns to display here.
One note is the Innkeeper tenants endpoint doesn't seem to return results sorted by create date the way other endpoints do? I'm default sorting in the table to get this order then. Probably better to filter in the store, but possibly should just change the innkeeper endpoint in Traction? Will get feedback from team and create minor story to adjust (or leave the table default sorted if we think that's fine)

image

**Check in**
Add tenant button/dialog/form.
Once checked-in, show the wallet id and key with a warning message (text up for feedback/revision). Once checked in, the parent dialog holder will get rid of the header X so you have to go through an "are you sure" button to close the dialog.

image

image

Need UX so this is a first cut, might change some later.

### DIFF
--- a/services/tenant-ui/frontend/src/components/innkeeper/tenants/CheckInTenant.vue
+++ b/services/tenant-ui/frontend/src/components/innkeeper/tenants/CheckInTenant.vue
@@ -1,0 +1,13 @@
+<template>
+    <div>
+        Hi
+    </div>
+</template>
+
+<script setup lang="ts">
+
+</script>
+
+<style scoped>
+
+</style>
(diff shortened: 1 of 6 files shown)


popkinj (approved):
This will make testing issuance so much easier. :)

usingtechnology (approved):
nice! 👍 👍 
only thing i can think of is maybe a copy to clipboard for wallet id/secret. could wait on that for how we want to deliver to a tenant...

PR #308, opened 2022-10-07
When Credential Template is in error, show that in the table

usingtechnology:

Added a StatusChip component and style sheet - build on this for all Statuses (there is a backlog ticket for that).


services/tenant-ui/frontend/src/assets/statusChip.scss
```diff
@@ -0,0 +1,16 @@
...
+  }
+
+  .p-chip.status-chip.in-progress {
+    background-color: lightsalmon;
+  }
+  
+  .p-chip.status-chip.error {
+    background-color: red;
+    color: white;
+  }
```
loneil, line 16: Was thinking this could go in the priemeVueComponents.scss file, then wer're not importing a file per control?
Though another option that is kind of clean would be one scss file per component under a /primevueComponents folder, then import all those into the priemeVueComponents.scss, then just import that one file into the main scss?

services/tenant-ui/frontend/src/assets/style.scss
```diff
@@ -1,4 +1,5 @@
 @import '_theme.scss'; // copied from '~primevue/resources/themes/nova/theme.css'
 @import 'layout.scss';
 @import 'forms.scss';
-@import 'primevueComponents.scss'
\ No newline at end of file
+@import 'primevueComponents.scss';
+@import 'statusChip.scss';
```
loneil, line 5: Could import this one in the one above so the top level styles file doesn't end up with all the specific primevue imports if we go that way

services/tenant-ui/frontend/src/assets/statusChip.scss
usingtechnology: yeah, not really sure what you wanted to do for organizing this stuff... maybe better to just have one file for tenant-ui components (after primevue?) I figured primevue was for overriding actual primevue classes only, and then file(s )for our components that are specific to tenant ui.

services/tenant-ui/frontend/src/assets/style.scss
usingtechnology: just let me know how you want to re-shuffle this stuff.

services/tenant-ui/frontend/src/assets/variables.scss
```diff
@@ -7,5 +7,18 @@ $tenant-ui-secondary-active: darken( $tenant-ui-secondary-color, 10% );
...
+$tenant-ui-status-green-background: #C8E6C9;
+$tenant-ui-status-amber-background: #FFD8B2;
+$tenant-ui-status-red-background: #FFCDD2;
+$tenant-ui-status-blue-background: #BCD7F1;
+
+$tenant-ui-status-default-color: #333333; // status chip text/colours
+$tenant-ui-status-green-color: #336C37;
+$tenant-ui-status-amber-color: #87623D;
+$tenant-ui-status-red-color: #C94040;
+$tenant-ui-status-blue-color: #485773;
```
loneil, line 20: Good stuff, we can use these vars for changing the toast to match mocks as well

loneil (approved):

PR #309, opened 2022-10-07
Better Issuer Button feedback 495

popkinj:
The Accept/Request issuer button now provides feedback if the user is approved.
1. There is a notification:
Screenshot from 2022-10-05 12-44-40
1. And the button deactivates and shows the user is currently an issuer:
Screenshot from 2022-10-05 12-20-09
This is all done without having to refresh the page.

Notes:
- The Issuer state was being disconnected when converting the store into refs.
- Lots of debugging to try and get Pinia to maintain state between the button and the store.
- Eventual solution was to create a local ref and update it by listening to mutations to the store.

### DIFF
--- a/services/tenant-ui/frontend/src/components/common/Issuance.vue
+++ b/services/tenant-ui/frontend/src/components/common/Issuance.vue
@@ -84,7 +84,7 @@ const label = () => {
     !pending.value
   ) {
     return 'Approve Issuer Permissions'; // Already approved, but not an issuer
-  } else if (pending) {
+  } else if (pending.value) {
     return 'Check Issuer Status'; // Requested, but not approved
   } else {
     return 'Request Issuer Permissions'; // Not approved yet
--- a/services/tenant-ui/frontend/src/components/common/Issuance.vue
+++ b/services/tenant-ui/frontend/src/components/common/Issuance.vue
@@ -1,9 +1,9 @@
 <template>
   <Button
     :label="label()"
-    :icon="tenant.issuer ? 'pi pi-check' : 'pi pi-flag'"
+    :icon="issuer ? 'pi pi-check' : 'pi pi-flag'"
     class="p-button"
-    :disabled="tenant.issuer"
+    :disabled="issuer"
     :loading="loading"
     @click="requestAccess"
   ></Button>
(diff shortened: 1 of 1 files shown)


loneil:
I don't seem to get the button updating after approving on my tenant

image

Navigating to and from that page doesn't have it change either, but if I go into the innkeeper UI and check the tenant it's innkeeper status is active. This is pointing against dev on my local

loneil:
If the issuer status is too asynchronous for the button to update right away at least we could add a "check back later" message to the toast maybe? Or an initial state flag that it's undetermined or something (might be overkill)
But if I do check back later (when the issuer state is active for sure) the button is still showing the approved state and I'd have to refresh so maybe it's not picking up the change there?

usingtechnology:
we don't have events going, so why don't we just use toast to say check back later (like @loneil says) and add a refresh button on that screen. seems to me we have to log out to get that profile data refreshed.

popkinj:
I'll see if I can duplicate what @loneil  is seeing. For me it's working great.... 😑
... and I like @usingtechnology idea of adding more info to the Toast message.

usingtechnology:
> I'll see if I can duplicate what @loneil is seeing. For me it's working great.... 😑
> ... and I like @usingtechnology idea of adding more info to the Toast message.

got it to work... little unintuitive? you have to hit the button... maybe reception the button to "refresh" or "check status"?

popkinj:
> 

Yeah... When you hit the button again if forces the app to check the profile status again.

popkinj:
I'm going to add another status to the button that indicates the request was sent and you can click the button again to check on the status.

popkinj:
OK... So here is the current behaviour.
If a profile has not be granted access to request issuer permissions they see this button:
Screenshot from 2022-10-11 10-00-22
If they click this button there is a request to the API and currently the response is declined. The following message is displayed:
Screenshot from 2022-10-11 10-00-07
If the request to the API takes a while the button looks like this:
Screenshot from 2022-10-05 12-10-54


If the profile has been granted access to request permissions they see this button:
Screenshot from 2022-10-05 12-19-34
When clicking this button they see the following message:
Screenshot from 2022-10-11 09-52-03
Then the button changes to this:
Screenshot from 2022-10-11 09-52-10
Clicking this button sends another call to the API to reload the users profile. When there is a positive flag for Issuer access the button turns into this:
Screenshot from 2022-10-05 12-20-09

Then the top right of the profile page looks like this:
Screenshot from 2022-10-11 09-52-22

usingtechnology (approved):
i think you can make it simpler without the issuer ref, but functionality works and that is the most important thing.
like the feedback on the button now... 👍 👍

PR #312, opened 2022-10-14
Generic Form Button - 507

popkinj:
Button component to standardize the look and behavior of all form buttons.
- When a label is supplied the button looks like this:
Screenshot from 2022-10-13 12-38-03
When there is no label it looks like this:
Screenshot from 2022-10-13 12-38-31
- An icon and function are mandatory props
- Disabled and loading states are optional props


loneil (changes requested):
See https://github.com/bcgov/traction/compare/develop...feature/genericFormPostStyle

PR #314, opened 2022-10-14
Localization structure

loneil:

Vue i18n structure. Don't add all translation locales yet (big job, mostly the translating part, leave to the community?).

Tried a few paths of this including implicit injection (true global)
https://vue-i18n.intlify.dev/guide/advanced/composition.html#implicit-with-injected-properties-and-functions
but note in that documentation kind of warns away from it. And Vue3 standards seem to be really moving away from globalProperties globals. So do the typical use composable like our other stuff (toast etc).

For the locale resources, we'll want to organize them when we start doing the actual translations. I don't know the best patterns for this stuff, like should it be per page? Per "concept"? Probably some master's dissertations on that stuff so I think we just go with what makes sense to us at this point.

Need a minor UI tweak to the changer but can leave as basic select box for right now.


Jsyro (approved):
Looks as expected to me. Awesome work.

loneil:
Sherman posted an example of how the BPA does locale file organization, https://github.com/hyperledger-labs/business-partner-agent/tree/main/frontend/src/locales

PR #316, opened 2022-10-21
Some Tenant UI things

loneil:
https://apps.nrs.gov.bc.ca/int/jira/browse/TRACTION-475
https://apps.nrs.gov.bc.ca/int/jira/browse/TRACTION-471
https://apps.nrs.gov.bc.ca/int/jira/browse/TRACTION-472

Revoke credential and delete credential in the issuer credentials list. I think this UI (action buttons that look similar) will need a revision but can go for now. 
image

Calls the revocation POST endpoint. Only shows the action when the cred is revokable (and not already revoked).
Delete calls the DELETE endpoint.

Package updates.

Other miscellaneous UI things.


popkinj (approved):
Nice work.

PR #322, opened 2022-10-30
Add i18n translations as locale json files

loneil:
Move message resources for localization to individual json files.

I discovered that our current build situation was using the tsc configuration for the toplevel of the app (IE the node backend) and transpiling the frontend each time it build the backend too. Think I set this up at the very beginning and we've been going with it so far.
This caused me an error when using a newer Vite-based import technique because the Node typescript setup can't do that esbuild stuff yet. So had to make a tweak to ignore the FE when building the BE. (then the FE is built separately as before with its own tsconfig)
Think there's some more work to do to have the tenant UI building be the best way it can be... so can discuss later on. But it works for now locally and with Helm.

Please review:
https://github.com/bcgov/traction/pull/322/files#diff-2d4fc84d8f16d51ce5e9c816b09c5588f1877bd8284b20356004e0530cd282bf
https://github.com/bcgov/traction/pull/322/files#diff-ebaa196949c5f7f225ce3bdbd650bb2312381d82611535c1cb80cd759bd0bc6d
https://github.com/bcgov/traction/pull/322/files#diff-ebaa196949c5f7f225ce3bdbd650bb2312381d82611535c1cb80cd759bd0bc6d
https://github.com/bcgov/traction/pull/322/files#diff-d4bac62e0c2bd8318ed18a8c9b3614ca90f56ecbad569864e656a8496a19c2f5

(can ignore the rest if you want as they're just more translation examples being thrown into Vue components).

### DIFF
--- a/services/tenant-ui/frontend/src/components/common/LocaleSwitcher.vue
+++ b/services/tenant-ui/frontend/src/components/common/LocaleSwitcher.vue
@@ -8,7 +8,7 @@ import { useI18n } from 'vue-i18n';
 
 const { locale } = useI18n({ useScope: 'global' });
 
-const localeList = ['en', 'fr'];
+const localeList = ['en', 'fr', 'jp'];
 </script>
 
 <style lang="scss" scoped>
--- a/services/tenant-ui/README.md
+++ b/services/tenant-ui/README.md
@@ -63,3 +63,15 @@ Build and run a docker image (example shows using environment variable to point 
 docker build . -t local/traction-ui
 docker run --env SERVER_TRACTION_URL=https://traction-api-test.apps.silver.devops.gov.bc.ca/ -p 8080:8080 -d local/traction-ui
 ```
+
+## Internationalization
+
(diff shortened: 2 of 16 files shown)


services/tenant-ui/frontend/src/components/common/LocaleSwitcher.vue
```diff
@@ -8,7 +8,7 @@ import { useI18n } from 'vue-i18n';
 
 const { locale } = useI18n({ useScope: 'global' });
 
-const localeList = ['en', 'fr'];
+const localeList = ['en', 'fr', 'jp'];
```
popkinj, line 11: Just curious... Is this Japanese? If so, do we need to support this?

popkinj (approved):
Looks good @loneil . Would love to see a walk through.

services/tenant-ui/frontend/src/components/common/LocaleSwitcher.vue
loneil: Yeah, it's just a sample language though, not going to do any of the translations at this point, wanted one to demo that has non English characters though.

PR #327, opened 2022-11-08
Feature/OIDC inkeeper

loneil:
https://apps.nrs.gov.bc.ca/int/jira/browse/TRACTION-545
Add configurable OIDC login for the frontend. Add secured backend functionality to login with the stored ,**secret**, innkeeper token and return that to the FE.

After getting the innkeeper token, the frontend functions as it did before, using that token to make calls to Traction. As such the OIDC JWT is only needed at the very beginning (at this point, may expand use later) so token renewal there is turned off.

The frontend will support using the admin key as before, a configured OIDC provider, or both, controlled by configuration.  
image

UX final design to come.

For this PR deployment (and any promotions) the Admin mode is enabled and OIDC is deactivated. I need to configure up some things in Keycloak, and add the appropriate config to helm deployments after that. So a subsequent quick PR/ticket will enable this in our OCP environments.

If you want to test this out locally, you can set up your `local.json` to set `frontend.oidc.active` to true and set the `server.innkeeper.key` to the Innkeeper key in the **environment you are using at the time for Traction**.

To be an innkeeper, contact me once you've logged in and I can grant the role in our DEV keycloak realm. Final role name (and if it's a realm role or not) TBD so some minor config might change in next PR too.

### DIFF
--- a/services/tenant-ui/.eslintrc
+++ b/services/tenant-ui/.eslintrc
@@ -4,6 +4,9 @@
   "plugins": [
     "@typescript-eslint"
   ],
+  "rules": {
+    "@typescript-eslint/no-explicit-any": "off",
+  },
   "extends": [
     "eslint:recommended",
     "plugin:@typescript-eslint/eslint-recommended",
--- a/services/tenant-ui/frontend/src/components/innkeeper/InnkeeperLoginOidc.vue
+++ b/services/tenant-ui/frontend/src/components/innkeeper/InnkeeperLoginOidc.vue
@@ -1,8 +1,11 @@
 <template>
-  <Button class="w-full mt-5" label="IDIR" @click="oidcLogin" :loading="loading"/>
-  <div v-if="error">
-    ERROR: {{ error }}
-  </div>
+  <Button
+    class="w-full mt-5"
+    label="IDIR"
+    :loading="loading"
+    @click="oidcLogin"
+  />
+  <div v-if="error">ERROR: {{ error }}</div>
 </template>
 
 <script setup lang="ts">
@@ -22,7 +25,6 @@ const oidcLogin = async () => {
   try {
     await innkeeperOidcStore.login();
   } catch (error: any) {
-    debugger;
     toast.error(`Failure: ${error}`);
   }
 };
--- a/services/tenant-ui/frontend/src/components/innkeeper/InnkeeperLoginOidc.vue
+++ b/services/tenant-ui/frontend/src/components/innkeeper/InnkeeperLoginOidc.vue
@@ -1,0 +1,26 @@
+<template>
+  {{ config }}
+  <Button @click="oidcLogin" class="w-full mt-5" label="IDIR" />
+
+  {{ user }}
+</template>
+
+<script setup lang="ts">
+//Vue
+import { ref, reactive } from 'vue';
+// State
+import { useConfigStore } from '@/store';
+import { useInnkeeperOidcStore } from '@/store';
+import { storeToRefs } from 'pinia';
+// PrimeVue/Validation/etc
+import Button from 'primevue/button';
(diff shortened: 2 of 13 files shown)


services/tenant-ui/.eslintrc
```diff
@@ -4,6 +4,9 @@
   "plugins": [
     "@typescript-eslint"
   ],
+  "rules": {
+    "@typescript-eslint/no-explicit-any": "off",
```
loneil, line 8: Had to add this because middleware injects new fields into REquest (common thing) and then TS doesn't like that... I'm sure there's a standard way to handle that we can look up.

services/tenant-ui/frontend/src/store/innkeeper/innkeeperOidcStore.ts
```diff
@@ -0,0 +1,111 @@
...
+  //   .catch((err) => {
+  //     debugger;
+  //     error.value = err;
+  //   });
+
+  // if (error.value != null) {
+  //   debugger;
+  //   // throw error so callers can handle appropriately
+  //   throw error.value;
+  // }
```
loneil, line 86: To take out

services/tenant-ui/frontend/src/components/innkeeper/InnkeeperLoginOidc.vue
```diff
@@ -0,0 +1,31 @@
+<template>
```
loneil, line 1: UX for this file possibly to change in Traction UX design tweaks epic.

services/tenant-ui/frontend/src/store/innkeeper/innkeeperOidcStore.ts
```diff
@@ -0,0 +1,111 @@
...
+import { useInnkeeperTokenStore } from './innkeeperTokenStore';
+import { UserManager } from 'oidc-client-ts';
+
+export const useInnkeeperOidcStore = defineStore('innkeeperOidcStore', () => {
+  // other stores
+  const { config } = storeToRefs(useConfigStore());
+  const { token } = storeToRefs(useInnkeeperTokenStore());
+
+  // private (move to other file maybe?)
+  const _settings: any = {
```
popkinj, line 14: Something tells me the linter will complain about variables/functions starting with an underscore.

services/tenant-ui/package.json
```diff
@@ -14,12 +14,15 @@
...
   "dependencies": {
+    "axios": "^0.27.2",
     "body-parser": "^1.20.1",
     "config": "^3.3.8",
     "connect-history-api-fallback": "^2.0.0",
     "cors": "^2.8.5",
     "express": "^4.18.2",
+    "express-oauth-jwt": "^2.0.1",
     "http-proxy-middleware": "^2.0.6",
+    "jose": "^4.10.4",
```
popkinj, line 25: Curious as to what this library does.

popkinj (approved):
Looks good. :+1:

services/tenant-ui/package.json
loneil: Jose is a JWT verification specification. Lots of libraries in lots of languages to support it. The Java one has a good explanation of where it comes from https://bitbucket.org/b_c/jose4j/wiki/Home

services/tenant-ui/frontend/src/store/innkeeper/innkeeperOidcStore.ts
loneil: Lint and code climate seem to be ok with it, pretty standard for "private" vars in JS I think.

usingtechnology (approved):
looks good to me.
we can add KC config to CI/CD later once it is decided how to leverage this.

PR #328, opened 2022-11-09
Force plugins to use acapy 0.7.4

usingtechnology:

One plugin was pulling in 0.7.5 which resulted in a payload error with an endorser running 0.7.4. Let's remain (and ensure we use) 0.7.4 in our plugins, thus our custom acapy agent image.


loneil (approved):
👍

PR #331, opened 2022-11-18
Feature/traction innkeeper plugin

usingtechnology:
PoC for moving innkeeper functionality into a plugin.
Allows creation and management of tenants (no real data or functionality added to tenant, so more of a placeholder for now).
Allows for prospective users of the service to register and get approved, then create a wallet (they will be responsible for storing their wallet_key).

Innkeeper will have it's own tenant/wallet and use that for security on API calls, and will allow for future improvements such as issuing credentials (ex. issue to new tenant owner to create their wallet and administer their tenant) and verification, and communication with staff (via basic messages to know connections)... that's waaaaaaaaaay down the road but having a tenant/wallet opens up a lot of functionality.

see the plugin readme for quick overview of the flow and endpoints.

Does not include helm chart configuration. Only docker and docker compose. Can put the helm configuration in this PR or make another one.


popkinj (approved):
The code changes don't mean much to me but  I like and can understand the Readme.

PR #332, opened 2022-11-18
Call acapy plugin from Tenant UI

loneil:
**Acapy calls from tenant UI**
Allow the Tenant UI frontend to proxy calls over to acapy plugins while still keeping Traction calls in place.
- Added a new proxy for routes from the FE that end with `/acapy`. This routing sends the requests to the acapy admin api instead of Traction.
- Calling this proxy injects the `X-API-KEY` in the tenant UI backend so the acapy admin calls can go through
- Because this api key is injected, Tenants are allowed to call acapy (we need this for plugins) but that key is indiscriminate (need to think on this for LOB users calling our plugins... discussed with Sherman). So there's an acapy middleware now that sets an allowed list of URLS. I think we'll have to live with this for now.
- Proxy gets the acapy JWT out of the Traction JWT and uses that for acapy calls.
- Nothing much to change on FE other than setting a different path (and handling a response slightly differently if it's acapy)

**Config/deployment**
Set up all the Tenant UI config and secret management so that the _backend_ of the Tenant UI has access to environmental config it needs for acapy calls.
- Add some Helm stuff so that the Tenant UI deployment can map to Traction/acapy deployment secrets

**Frontend**
Cut the messages list over to call acapy basicmessaging plugin, add a Create Message UI that looks up the contact (and get acapy connection ID) and a textbox for content
- Lost a couple things with the changes. No row expander because there's no individual message endpoint. Not sure we'll need this going forward? UX dependent.
- Can't show the actual contact name from this endpoint call. Maybe something to discuss/enhance shortly. Could do the ID->contact mapping on the FE if we want?
- The "Sent" date is not in order, will see if we need to handle that api side or client side. (easy enough in the table if we want)

**Developer page**
Needed to troubleshoot config so added a developer page. Adding a ticket to enhance this to be configurable to show or not and add other stuff to there.
Also at this point FE config shows server config (secrets!!!!) so that can't happen. Addressing in my other ticket in this sprint but it's just lower env for now.


popkinj (approved):
Lots of good stuff here. I appreciate all the comments.
Nice work @loneil .

charts/tenant-ui/values-test.yaml
```diff
@@ -9,4 +9,6 @@ tenant_ui:
     pullPolicy: Always
   traction_api:
     endpoint: https://traction-api-test.apps.silver.devops.gov.bc.ca
+  acapy_admin_api:
+    endpoint: https://traction-acapy-admin-test.apps.silver.devops.gov.bc.ca/
```
usingtechnology, line 13: does trailing slash matter?

charts/tenant-ui/values-int.yaml
```diff
@@ -9,4 +10,6 @@ tenant_ui:
     pullPolicy: Always
   traction_api:
     endpoint: https://int-traction-api-test.apps.silver.devops.gov.bc.ca
+  acapy_admin_api:
+    endpoint: https://int-traction-acapy-admin-test.apps.silver.devops.gov.bc.ca/
```
usingtechnology, line 14: does trailing slash matter? see values-test too...

services/tenant-ui/src/index.ts
```diff
@@ -38,6 +44,11 @@ app.use("/config", (_, res, next) => {
 // This service's api endpoints
 app.use(APIROOT, router);
 
+// Proxy any api/traction/acapy calls over to acapy
+// This is just to support developing the FE against both paths for now
+// See comments in acapy.ts middleware
+app.use(`${PROXYACAPYROOT}/`, handleAcapy, acapyProxy);
```
usingtechnology, line 50: smart... nice separation

usingtechnology (approved):
nice stuff.
we can just keep an eye on the deployment flow to dev -> test -> int in case some of the naming or substitutions are off. super hard (impossible?) to test without doing it .

charts/tenant-ui/values-test.yaml
loneil: Good catch, it probably would in int and test, removing

PR #334, opened 2022-11-24
basicmessages_storage - list sort by created_at desc

usingtechnology:

Make default sort be on `created-at` descending, do not factor in `state` or `connection_id`


loneil (approved):
Looking good

PR #335, opened 2022-11-25
Contact name in the messages table

popkinj:
Signed-off-by: Jamie Popkin <popkinj@littleearth.ca>
Screenshot from 2022-11-24 14-09-46
- Contact name has replaced the ID
- API schema generation logic


popkinj:
Hmmm... Should I be changing "Contact" to "Connection"?

loneil:
Would we need a contact load in the messages component? I haven't tried out yet but I think this wouldn't be able to map if the contacts weren't pre-loaded?
If so could add it as part of the load table function

popkinj:
> Would we need a contact load in the messages component? I haven't tried out yet but I think this wouldn't be able to map if the contacts weren't pre-loaded? If so could add it as part of the load table function

I'll give it a test. I thought just accessing the store causes it to request the data... will find out.

loneil:
> > Would we need a contact load in the messages component? I haven't tried out yet but I think this wouldn't be able to map if the contacts weren't pre-loaded? If so could add it as part of the load table function
> 
> I'll give it a test. I thought just accessing the store causes it to request the data... will find out.

No accessing state from a store doesn't do a fetch or anything. The action must be called to populate the state. So if you just refresh on the messages page, or log in without first going to contacts and then go to messages the contacts list won't be in state.

image

popkinj:
> > > Would we need a contact load in the messages component? I haven't tried out yet but I think this wouldn't be able to map if the contacts weren't pre-loaded? If so could add it as part of the load table function
> > 
> > 
> > I'll give it a test. I thought just accessing the store causes it to request the data... will find out.
> 
> No accessing state from a store doesn't do a fetch or anything. The action must be called to populate the state. So if you just refresh on the messages page, or log in without first going to contacts and then go to messages the contacts list won't be in state.
> 
> image

OK... nice catch @loneil

loneil (approved):
Looks like it's working as expected

PR #338, opened 2022-11-28
UI Modifications

loneil:
Round of changes from 
https://apps.nrs.gov.bc.ca/int/jira/browse/TRACTION-572
https://apps.nrs.gov.bc.ca/int/jira/browse/TRACTION-588

A bit of the toasting style might not be finalized and need some input from Guru but that will be for later


popkinj (approved):
:+1:

PR #339, opened 2022-11-30
Feature/tenant proxy

usingtechnology:
For "direct" access the acapy api, we need to add a proxy that does not require the X-API-ADMIN key, only the tenant/wallet's token.

Use NGINX to expose this limited access - deny access to acapy endpoints that a tenant does not need and should not access.

This builds a new nginx image with our configuration, takes some environment variables to point at the right acapy and use the configure acapy admin URL key.

Update github actions to build the new image.

Put in helm charts to deploy.


loneil:
Can uninstall the `jsonwebtoken` library (and `"@types/jsonwebtoken`) it looks on the backend of tenant UI now if the key token isn't being extracted in that any more

usingtechnology:
> Can uninstall the `jsonwebtoken` library (and `"@types/jsonwebtoken`) it looks on the backend of tenant UI now if the key token isn't being extracted in that any more

got it. making that change now.

loneil (approved):

PR #340, opened 2022-12-01
Table general styling

loneil:
For https://apps.nrs.gov.bc.ca/int/jira/browse/TRACTION-574

image


services/tenant-ui/frontend/src/helpers/constants.ts
```diff
@@ -0,0 +1,4 @@
+export const TABLE_OPT = {
```
usingtechnology, line 1: constants! let's go!!

usingtechnology (approved):
great stuff. looking good.

PR #341, opened 2022-12-02
Move API Paths into constants.

usingtechnology:

Moving all API paths into the constants file. This should help refactoring as we move forward. There are some constants that are currently not used but exist in the APIs. 

Note that the complete traction API is not in constants.

Also, note that some "constants" are functions. These are the ones where an ID is present in the path.

### DIFF
--- a/services/tenant-ui/frontend/src/store/innkeeper/innkeeperOidcStore.ts
+++ b/services/tenant-ui/frontend/src/store/innkeeper/innkeeperOidcStore.ts
@@ -15,10 +15,10 @@ export const useInnkeeperOidcStore = defineStore('innkeeperOidcStore', () => {
   const _settings: any = {
     authority: config.value.frontend.oidc.authority,
     client_id: config.value.frontend.oidc.client,
-    redirect_uri: API_PATH.OIDC_INNKEEPER_REDIRECT_URI,
+    redirect_uri: `${window.location.origin}/innkeeper`,
     response_type: 'code',
     automaticSilentRenew: false, // don't need to renew for our needs at this point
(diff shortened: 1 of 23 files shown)


services/tenant-ui/frontend/src/store/innkeeper/innkeeperOidcStore.ts
```diff
@@ -14,10 +15,10 @@ export const useInnkeeperOidcStore = defineStore('innkeeperOidcStore', () => {
   const _settings: any = {
     authority: config.value.frontend.oidc.authority,
     client_id: config.value.frontend.oidc.client,
-    redirect_uri: `${window.location.origin}/innkeeper`,
+    redirect_uri: API_PATH.OIDC_INNKEEPER_REDIRECT_URI,
     response_type: 'code',
     automaticSilentRenew: false, // don't need to renew for our needs at this point
-    post_logout_redirect_uri: `${window.location.origin}/innkeeper`,
```
loneil, line 20: I'd lean towards leaving window location code out of constants, just for reading the code and knowing exactly what it's doing in the store here... but that's just me, could go either way

loneil (approved):
One comment, can take or leave the change. Tested through innkeeper and tenant (Traction and ACAPy calls) on the deployed PR. Looks good to me

services/tenant-ui/frontend/src/store/innkeeper/innkeeperOidcStore.ts
usingtechnology: fair enough. i just started moving everything over but was on the fence about this. I'll move it out since you noticed it too.

PR #342, opened 2022-12-06
Feature/reservation form

loneil:
Simple first step of reservation process, requesting a tenant.

For this PR, just implement the form itself, which calls the proxy /reservation endpoint directly. See some comments in the code about the store/API call.

For this, just do a success toast when the response is good. Next steps about what to show in the response on the next page coming in the next ticket.

image

image


plugins/docker/tenant-proxy.conf.template
```diff
@@ -68,6 +68,43 @@ server {
         }
     }
 
+    location ~ ^/multitenancy/reservations$ {
```
loneil, line 71: Can adjust this with upcoming proxy settings refactor

services/tenant-ui/frontend/src/assets/style.scss
```diff
@@ -4,3 +4,34 @@
 @import 'tenantuiComponents.scss';
 @import 'primevueComponents.scss';
 @import 'toast.scss';
+
+// General HTML
```
loneil, line 8: Moving this out of an extraneous css file

services/tenant-ui/frontend/src/store/reservationStore.ts
```diff
@@ -0,0 +1,59 @@
...
+  const error: any = ref(null);
+  const reservation: any = ref(null);
+
+  // actions
+  async function makeReservation(payload: any = {}) {
+    console.log('> reservationStore.makeReservation');
+    error.value = null;
+    loading.value = true;
+    console.log(payload);
+    await api
```
loneil, line 26: Can't use the API store that all the other calls use since that results in a blank Bearer header from the interceptor and that 401s you.

services/tenant-ui/frontend/src/components/reservation/Reserve.vue
```diff
@@ -0,0 +1,171 @@
...
+const submitted = ref(false);
+const handleSubmit = async (isFormValid: boolean) => {
+  submitted.value = true;
+
+  if (!isFormValid) {
+    return;
+  }
+  try {
+    await reservationStore.makeReservation(formFields);
+    toast.success(`Your request was recieved.`);
```
usingtechnology, line 165: i realize this is short-term but spelling: "received"

services/tenant-ui/frontend/src/store/reservationStore.ts
usingtechnology: ahhh, nice.

usingtechnology (approved):
looks good, deal with spelling or not, I'm sure that toast is going soon anyway.

PR #346, opened 2022-12-08
Handle self-message loop bug by not listening for the event.

usingtechnology:
Deprecate the traction basic message api.
Leave endpoints there, return "Sunset" and "Deprecated" headers (as suggested [here](https://blog.stoplight.io/deprecating-api-endpoints).
Have discussed this with the team, and we will return `410 Gone` status.

I have left the code behind for reference purposes (and if we had to turn it back on) but could be talked into removing the code completely.

Setting to return a default ack message using acapy flag `--auto-response-messages`

### DIFF
--- a/charts/traction/values.yaml
+++ b/charts/traction/values.yaml
@@ -138,7 +138,7 @@ acapy:
   staticArgs:
     autoAcceptInvites: true
     autoAcceptRequests: true
-    autoRespondMessages: false
+    autoRespondMessages: true
     autoRespondCredentialProposal: false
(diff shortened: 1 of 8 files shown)


usingtechnology:
Screen Shot 2022-12-08 at 12 08 04 PM

Deprecated API in swagger

loneil (approved):
If we're good with losing the auto-response message for now (which I think we are?) then it looks good to me

popkinj (approved):
Looks good. :)

PR #347, opened 2022-12-09
Allow prospective tenant to check state of their reservation.

usingtechnology:
Allow innkeeper to deny request and provide a reason.


loneil (approved):
Looks good to me

PR #353, opened 2022-12-09
Innkeeper Title

popkinj:
- The default page title get's overridden with a _Router Guard_ for all innkeeper pages.
- Small spelling fix.


services/tenant-ui/frontend/src/router/index.ts
```diff
@@ -21,4 +21,16 @@ const router = createRouter({
...
 
+/**
+ * Global router guard to set the page title.
+ * Override only if in the non-default tenant UI.
+ * In this case, the innkeeper UI is the non-default.
+ * TBD: Could move this into a Helm chart value.
+ */
+router.afterEach((to) => {
+  if (to.meta.isInnkeeper === true) {
+    document.title = 'Traction Innkeeper Console';
```
usingtechnology, line 32: let's make this an env var -> config like the tenant side.
see and `UX_APP_TITLE` and `document.title = config.value.frontend.ux.appTitle;`

we can split the work, if you get it into the application code and the scripts/docker-compose. i can run with getting the helm charts and CICD done

usingtechnology (changes requested):

services/tenant-ui/frontend/src/views/innkeeper/InnkeeperTenants.vue
```diff
@@ -4,4 +4,7 @@
 
 <script setup lang="ts">
 import Tenants from '@/components/innkeeper/tenants/Tenants.vue';
+
+// Override the page title. KISS.
+document.title = 'Traction Innkeeper Console';
```
loneil, line 9: Is this needed as well as the router guard one above? Repeating the same string.

popkinj: No... totally not. Thought I removed it.
Good catch @loneil

services/tenant-ui/frontend/src/router/index.ts
popkinj: OK.... Think I've got this working.
I tried being ambitions and updated one of the Helm files..... Caused the pipeline to explode.... Changed it back. I'll let @usingtechnology do the Helm stuff.

scripts/docker-compose.yml
```diff
@@ -220,6 +220,7 @@ services:
       - IMAGE_TAG=${IMAGE_TAG}
       - IMAGE_VERSION=${IMAGE_VERSION}
       - UX_APP_TITLE=${UX_APP_TITLE}
+      - UX_APP_INNKEEPER_TITLE=${UX_APP_INNKEEPER_TITLE}
```
loneil, line 223: Can add this one to .env-example under UX_APP_TITLE (line 176) to pull in to docker compose.

usingtechnology (approved):

PR #358, opened 2022-12-14
Feature/askar db per wallet

usingtechnology:
Change from indy wallets to askar wallets.
Need to update endorser to askar (other PR).
Need to update our local dev scripts and our helm charts.


loneil (approved):
I think it's good. Did the following through the tenant UI at least

- created reservation
- log in through OIDC to innkeeper 
- created tenant
- log in with tenant
- become issuer
- create schema
- make cred
- connect to BC wallet
- message back and forth
- issue cred
- revoke cred

PR #360, opened 2022-12-16
Approve and deny reservation

loneil:
Actions to approve and deny reservations in the innkeeper UI.
Some UX questions coming out of this so may be subject to change in the future a bit.
The approve returns the PW to the screen right now for 2 reasons
1 Email SMTP connection still to come
2 I think we need that as an option anyways since email might be optional and it could fail.

image

image

image

image

image


usingtechnology (approved):
looks good to me. obviously some short term things in place so let's keep rolling. makes sense to not put too much in one PR.

PR #361, opened 2022-12-17
Use proper return code (404) when record not found.

usingtechnology:


loneil (approved):
Looks good. pulled locally and poked around the innkeeper calls

PR #362, opened 2022-12-20
Send emails for reservation flow through Node API

loneil:
Send emails when
1. The user makes a reservation
-- email to the user themself
-- email to the innkeeper

1. The Innkeeper accepts or declines the invitation
-- email is sent back to requestor with **password** in it if accepted

Emails are done in the Node API backend of the Tenant UI using nodemailer and whatever smtp server you want to configure. Remember if using gov to be on the VPN if doing locally.

Add configuration to set the From email, as well as the innkeeper inbox. **All this still needs Helm values** but we can get it in with the FE defaults for now

The emails themselves are placeholders. We need UX flows for email contents and wordings to come so will be working with Kyle and Guru on that.

Right now the emails are being sent out from separate calls from the frontend that just kick off in non-blocking calls and they will fail silently if the api doesn't return success. I still need to sort out some requirements about the optionality of using emails since the UX designs have the password delivery done in email as a necessity (though I am ALSO returning the PW to the screen at this point). I _think_ what we'll want to do is just not send emails if there is no SMTP configured, or a 'use email' flag or something. Then decide if we would want to return errors in the event you are relying on email. Not sure, still need to sort some requirements but this all works well for now.


usingtechnology (approved):
looks good for first round.

PR #364, opened 2022-12-24
Bugfix/wallet key verify

usingtechnology:
big changes to the multitenant_provider plugin.

first, made it `v1_0` to match our other plugins.

most important change:

allow `unmanaged` wallets, and provide a check on (unmanaged wallets) asking for a token - does the `wallet_key` passed in match the saved key?

note: "saved" keys are just salted hashes of the original key used to create the wallet and returned to the owner. we do not save the actual key for `unmanaged` wallets.

changes to the configuration of this plugin, allow setting the expiry date of the token, and a flag to silently allow `managed` wallets to call the `multitenancy/wallet/<wallet id>/token` endpoint with a `wallet_key` - the key is either ignored or raises an error (base acapy and the plugin default behaviour).

updates to the traction_innkeeper plugin to create `unmanaged` wallets for new tenants. 

updates to `scripts` and `charts` to pass in all the new configuration.


loneil (approved):
Think it's all making sense to me

PR #365, opened 2022-12-28
Reservation status and check-in

loneil:


popkinj (commented):
Looks good. Like that you found a replacement for the embedded svg.

loneil:
> Looks good. Like that you found a replacement for the embedded svg.

No need for images when they're primevue icons.

popkinj:
> 

I used the svg because it was the closest thing I could find to the mockup... But I think it's probably better to stick with our existing fontset.

loneil:
> > 
> 
> I used the svg because it was the closest thing I could find to the mockup... But I think it's probably better to stick with our existing fontset.

The icon images in the mockups are primevue icons. We use those so we don't have to add images. The one used by Guru here https://xd.adobe.com/view/04d57e67-4b0a-49c7-bf15-ea74939aa66e-5186/screen/460d9770-1a22-449a-a3ed-d5ab95db5917/ is the `pi-check-circle` from the iconset https://www.primefaces.org/primevue/icons

PR #376, opened 2023-01-06
Add new endpoint to read a connection's invitation

usingtechnology:

Added the new handler in a way that acts like a separate plugin.


popkinj (approved):
Looks good.

loneil (approved):
Think it looks good. I can make and get invites with tokens from my tenant. 
I can't get connections with tokens from a different tenant (404). So 👍

PR #383, opened 2023-01-12
Invitations

loneil:
Add in the connections and invitations tables. Both of these are just the results of the "connections" list from acapy, but we filter by whether they are in invitation state to determine which table has which records.

Single and multi-use buttons added and called from Invitations table.

Still things to do with this in that we want to invoke the new `connections/{conn_id}/invitation` endpoint to be able to go to a connection and re-fetch its invitation details. Mostly this will be to serve the ability to get the QR code/url for a multi-use invitation again after it's already been used.

Single-use invitations act like this:
- Invitation connection with mode "once" is created
- When actioned the status updates from invitation to active (or other statuses)
- Connection no longer has an invitation status so goes to connections table.

Multi-use invitations do the following (this is just acapy behaviour)
- Create a connection with mode "multi" (rather than "once") which is the invitation
- **Each** time that connection is actioned create a new connection to represent it
- This connection ends up with "once" as the mode, and a blank alias (need to think on what we want to do with this)
- The initial invitation stays around in the invitations list to be used again

Some of the file/folder names are pretty inconsistent with "contact" vs "connection" but I don't want to change that in case we switch up the concept naming again. Will do a pass over code naming once it's locked down.

image

image

### DIFF
--- a/services/tenant-ui/frontend/src/views/connections/MyInvitations.vue
+++ b/services/tenant-ui/frontend/src/views/connections/MyInvitations.vue
@@ -1,0 +1,7 @@
+<template>
+  <Contacts />
+</template>
+
+<script setup lang="ts">
+import Contacts from '../../components/contacts/Contacts.vue';
+</script>
--- a/services/tenant-ui/frontend/src/views/connections/MyInvitations.vue
+++ b/services/tenant-ui/frontend/src/views/connections/MyInvitations.vue
@@ -3,5 +3,5 @@
 </template>
 
 <script setup lang="ts">
-import Invitations from '../../components/invitations/Invitations.vue';
+import Invitations from '@/components/invitations/Invitations.vue';
 </script>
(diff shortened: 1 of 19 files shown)


services/tenant-ui/frontend/src/views/connections/MyInvitations.vue
```diff
@@ -0,0 +1,7 @@
+<template>
+  <Invitations />
+</template>
+
+<script setup lang="ts">
+import Invitations from '../../components/invitations/Invitations.vue';
```
popkinj, line 6: May be good to start using "@" instead of relative paths.

popkinj (approved):
Looks good @loneil

services/tenant-ui/frontend/src/views/connections/MyInvitations.vue
loneil: 👍

PR #384, opened 2023-01-13
Need endorser for Public DID and Issuer

usingtechnology:

1. Add endorser config back in to CICD (use old/existing endorser service).
2. Add endorser back into local dev scripts (use old endorser service code).
3. Add endpoint to support out of band register of DID

There are tickets added in JIRA documenting how to get connected to endorser (code added in plugin to add connection metadata) and how to get a public DID (endpoint added to plugins to support that flow).

Future should use the external endorser services code. https://github.com/hyperledger/aries-endorser-service


loneil (approved):
Haven't had a chance to put it through too many paces but I think it looks good so far

PR #385, opened 2023-01-14
Regenerate invitation

loneil:

Add an action to the invitations table to re-get the invite
image

Brings up the QR code and the link (just like the create screen)
image

Uses the fetch invitation endpoint with the connection ID, since the invitations are themselves just connections in invitation state
https://traction-tenant-proxy-dev.apps.silver.devops.gov.bc.ca/api/doc#/connection/get_connections__conn_id__invitation


usingtechnology (approved):
nice work!
worked for me

PR #386, opened 2023-01-14
Hamburger

popkinj:
- By default the sidebar shrinks and expands depending on screen width
- If the user clicks the hamburger button the sidebar toggles to either narrow or wide.
- If the user clicks the hamburger button it is assumed they want the sidebar to stay in the desired state no matter what the screen width.


services/tenant-ui/frontend/src/store/stateStore.ts
```diff
@@ -0,0 +1,18 @@
+/**
+ * Place for storing global state that is not persisted in the backend.
+ */
+import { defineStore } from 'pinia';
+import { ref } from 'vue';
+
+export const useGlobalStateStore = defineStore('globalState', () => {
+  // state
+  const sidebarOpen: any = ref(null);
```
loneil, line 9: Should this be a true/false for components to know if the sidebar is open? If so I'd make it a boolean type and default it to false.
Or is it a string representing the class name? In which case I'd make it called sidebarOpenClass, type string, and default to whatever the starting class name is (or empty string if it's meant to be indeterminate)

services/tenant-ui/frontend/src/store/stateStore.ts
```diff
@@ -0,0 +1,18 @@
+/**
+ * Place for storing global state that is not persisted in the backend.
+ */
+import { defineStore } from 'pinia';
+import { ref } from 'vue';
+
+export const useGlobalStateStore = defineStore('globalState', () => {
```
loneil, line 7: All the stores are global and represent state so maybe a different naming here? "Common" or "Layout" or something?

services/tenant-ui/frontend/src/components/layout/Sidebar.vue
```diff
@@ -2,6 +2,7 @@
   <div class="traction-sidebar">
     <!-- <h1 v-if="tenant" class="sidebar-app-title">{{ tenant.name }}</h1> -->
     <h1 class="sidebar-app-title">Tenant UI</h1>
+    <h1 class="sidebar-app-title small">T</h1>
```
loneil, line 5: This is a placeholder so it might not be a T, will shortly be putting in the /tenant call to load the name for this header.
Probably just leave for now and I can put the logic in to take the first name as part of a computed property when that happens. Unless you know some CSS that would just take the first letter of whatever content or something.

services/tenant-ui/frontend/src/components/layout/AppLayout.vue
```diff
@@ -30,4 +30,95 @@ import Card from 'primevue/card';
...
+  if (sidebarOpen.value === null) {
+    // Use media queries
+    return null;
+  } else if (sidebarOpen.value) {
+    // Default width
+    return 'open';
+  } else {
+    return 'closed'; // Mobile width
+  }
+};
```
loneil, line 50: See comment in store about boolean or string.

services/tenant-ui/frontend/src/components/layout/AppLayout.vue
```diff
@@ -30,4 +30,95 @@ import Card from 'primevue/card';
...
+  min-width: '';
+  width: '';
+  @include open;
+}
+
+.layout-sidebar {
+  transition: width 0.2s ease-in-out;
+  transition: min-width 0.2s ease-in-out;
+}
+</style>
```
loneil, line 124: There's a layout.scss file that controls the other stylings from this component. This could go in there.

loneil (commented):
Not sure what the plan for the dropdowns are
image

don't think the blank items can be used. Does Guru have any plan for how those should look in the mocks? Could just add icons to them perhaps.

loneil:
Innkeeper will need it as well. The innkeeperLayout is a separate file. 

(Originally these were split in case they differed but haven't so far so the innkeeperLayout might be able to be swapped with appLayout with an isInnkeeper parameter that sets the 1 class that makes them different.)

services/tenant-ui/frontend/src/components/layout/Sidebar.vue
popkinj: Hmmm.... You can specify content in css depending on width. But I don't there there's a way to say just use first letter.

services/tenant-ui/frontend/src/store/stateStore.ts
popkinj: We kinda need three states here. That's why I'm using 'null' as a start point. Just plain boolean isn't flexible enough.
We could switch to using a string if you'd like to avoid `null | boolean`.

services/tenant-ui/frontend/src/store/stateStore.ts
popkinj: I like your thinking here.

services/tenant-ui/frontend/src/components/layout/AppLayout.vue
popkinj: That may be a good spot especially if the Innkeeper is going to look the same.

popkinj:
Screenshot from 2023-01-24 11-13-23

GurcharanjeetSingh:
Good to go ahead from the UX end.

loneil:
I like it!👍 
We may want to adjust breakpoint widths for when the sidebar collapses to coordinate with the flexbox grid, so tables and stuff would collapse at the same breakpoint. But we'll probably want to test around to see what the best sizes are for that stuff later.

loneil (approved):
I like it!👍
We may want to adjust breakpoint widths for when the sidebar collapses to coordinate with the flexbox grid, so tables and stuff would collapse at the same breakpoint. But we'll probably want to test around to see what the best sizes are for that stuff later.

PR #425, opened 2023-01-26
Set up tenant as Issuer

loneil:
Tenant Issuer self-serve setup.
On profile page
- Check endorser setup
- Create endorser connection 
- Register Public DID
- Display issuer status

Additional changes
- Load /self for Tenant on login
- Display tenant name in sidebar
- Profile Form with tenant details
- Delete connection

Put back Profile page and remove existing Make Issuer functionality that was there to use Endorser connection and Public DID registration with proper loading state and async handling.

**Public DID registration steps**
The Public DID registration does a lot of steps at once with some waits in there. This seems to work but we might want to discuss if this should be multiple actions rather than one? Can go with this for now but if there's problems we can break it down more I think.
Look for **registerPublicDid** in tenantStore.ts

Note that UX has some new things since designs so might have some changes to come.
I'm not sure about the 'Details' accordions but will leave for now as it's useful while building features.
image

An Endorser must be set up (check new endorser info endpoint from plugin). If there isn't you can't proceed

image

First you must connect to the endorser. DID registration is disabled until.

image

Connected **and** connection is active

image

In the event the connection takes too long and isn't active yet, will display a message

image

Once endorser is connected the button will do the steps to
- Create the DID
- Register it
- Assign it as the public DID
- Fetch the public DID

But it all looks like a single action (with a loader spinning) to the user.
image

image

Once everything is cleaned up the Issuer badge shows

image

**Connections**

The Endorser connection will show up in the list (we could _not_ do this) but if the connection matches the one from Endorser Info it won't let you delete it

image


popkinj (approved):
Looks good to me. Nice work @loneil

services/tenant-ui/frontend/src/store/tenantStore.ts
```diff
@@ -62,105 +66,183 @@ export const useTenantStore = defineStore('tenant', () => {
...
     if (error.value != null) {
       // throw error so $onAction.onError listeners can add their own handler
       throw error.value;
     }
     // return data so $onAction.after listeners can add their own handler
-    return tenantConfig.value;
+    return publicDid.value;
+  }
+
+  async function registerPublicDid() {
```
usingtechnology, line 175: This is great. Only thing I can think of (not necessary right now) is to break this up in the service/store, and have the "client" vue file call this sequence with pauses and allow it to update the screen about which step it is on. Not sure if we have any facility in the spinner to provide "live" feedback.

The flow and screens look good to me.

usingtechnology (approved):
great stuff.

services/tenant-ui/frontend/src/store/tenantStore.ts
loneil: Yeah breaking it up into discrete user actions is a UX (and error handling) decision we can iterate on.
Great idea about the screen text @usingtechnology , I've updated that like below for the multi-stage registration one

image

commit:  https://github.com/bcgov/traction/pull/425/commits/39469fc4afa77fab4cf4c0bc6842d408e350b482

PR #439, opened 2023-02-03
Tenant managed webhooks.

usingtechnology:
API matches `GET` and `PUT` `/multitenancy/wallet/{wallet_id}` but under `/tenant/wallet` - so once logged in tenants can manage their own webhooks and label.  The multitenancy API was for admin use only and required the `x-api-key` only - no bearer tokens.

Also included is eliminating the requirement for a unique tenant name. The provided tenant name will become the default agent/wallet label and will be used for the wallet name.

Only the wallet name needs to be unique, so if the provided name is NOT, we just add `-N` until it is unique. ie `fred-5`.

### DIFF
--- a/plugins/traction_innkeeper/traction_innkeeper/v1_0/tenant/routes.py
+++ b/plugins/traction_innkeeper/traction_innkeeper/v1_0/tenant/routes.py
@@ -1,0 +1,223 @@
+import logging
+
+import requests
+from aiohttp import web
+from aiohttp_apispec import (
+    docs,
+    response_schema,
+    querystring_schema,
+    request_schema,
+)
+from aries_cloudagent.admin.request_context import AdminRequestContext
+from aries_cloudagent.messaging.models.openapi import OpenAPISchema
+from aries_cloudagent.messaging.valid import (
+    INDY_DID,
+    INDY_RAW_PUBLIC_KEY,
+)
+from aries_cloudagent.multitenant.admin.routes import (
+    format_wallet_record,
+    UpdateWalletRequestSchema,
+)
+from aries_cloudagent.multitenant.base import BaseMultitenantManager
+from aries_cloudagent.wallet.base import BaseWallet
+from aries_cloudagent.wallet.error import WalletError
+from aries_cloudagent.wallet.models.wallet_record import (
+    WalletRecordSchema,
+    WalletRecord,
+)
+
+from marshmallow import fields
+
+from ..routes import error_handler
+from ..tenant_manager import TenantManager
+from ..models import (
+    TenantRecord,
+    TenantRecordSchema,
+)
+
+LOGGER = logging.getLogger(__name__)
+
+SWAGGER_CATEGORY = "traction-tenant"
+
+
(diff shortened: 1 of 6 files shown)


loneil (approved):
Tested out creating a few tenants, successfully added the "-N" part at the end. Might be some UX design decisions (telling people etc) for the Tenant UI down the line but can check on that later.

Noticed it's not case sensitive, which I guess means the Acapy unique constraint is **not** case sensitive, since it worked... As discussed I think this is ok but can see if there's operational preferences and such to normalize all case or something

image
 
image

image

Tenant settings endpoint looks good to me in swagger, I could update them on a tenant on this PR.

Will have to do some tweaking on the Tenant UI to use that label maybe... right now it's displaying name from the /tenant endpoint... We'd probably want to use label, but also maybe show them name somewhere, can think it through UX wise.

esune (commented):
The `GET` and `PUT` payloads under the `tenant/wallet` namespace will not return the same set of data: in particular, the webhook URL seems to be set with the `PUT`, but there is no way to fetch the current value.

plugins/traction_innkeeper/traction_innkeeper/v1_0/tenant/routes.py
```diff
@@ -0,0 +1,223 @@
...
+    context: AdminRequestContext = request["context"]
+    # we need the caller's wallet id
+    wallet_id = context.profile.settings.get("wallet.id")
+
+    # records are under base/root profile, use Tenant Manager profile
+    mgr = context.inject(TenantManager)
+    profile = mgr.profile
+
+    # this is from multitenant / admin / routes.py -> wallet_get
+    # duplicate code.
```
esune, line 142: Would it make sense (and would it be possible), if the code is the same, to just trigger the original handler and return its result, rather than duplicating the code here?

usingtechnology: don't think it's possible as you will need to manipulate the request itself to get into the call...(need to put the wallet_id into the request URL) could redirect? but that seems kind of weird too.

usingtechnology:
> The `GET` and `PUT` payloads under the `tenant/wallet` namespace will not return the same set of data: in particular, the webhook URL seems to be set with the `PUT`, but there is no way to fetch the current value.

it just follows whatever the acapy calls are doing.  the PUT call is a totally different structure than the result - but that is how Aca-Py behaves. So this follows the API exactly, except a tenant calls it instead of some administrator on their behalf.

esune:
> > The `GET` and `PUT` payloads under the `tenant/wallet` namespace will not return the same set of data: in particular, the webhook URL seems to be set with the `PUT`, but there is no way to fetch the current value.
> 
> it just follows whatever the acapy calls are doing. the PUT call is a totally different structure than the result - but that is how Aca-Py behaves. So this follows the API exactly, except a tenant calls it instead of some administrator on their behalf.

Talked to @usingtechnology: the swagger doc doesn't expose the types contained in the `settings` property, which is where the webhook urls and other settings are returned.

esune (approved):

PR #438, opened 2023-02-03
Schema Storage usage in Tenant UI

loneil:
Adjust schema page with other additional improvements to call plugin schema storage for schema list/creation/import.
Only issuer tenants can create, others can copy/import.

Does not do any of the ledger listing, or schema-sync stuff from the storage plugin. Might want to discuss places to fit that stuff in but this can allow us to move forward with issuing flow at this point.

Some asynchronicity with storage appearing after creating so for now just warning the user they may need to refresh to see it appear in table.

If anyone wants to try on this PR without setting up tenants these ephemeral ones can be used

Issuer
806ba96b-6548-4972-a254-1995ac334ddb
021f52c9-5632-4f75-a214-3f33ceb4fce5

Non-Issuer
d586d345-0694-4fc5-8ca3-3632e6c8eb9e
e5798ce7-ad12-4b0d-ab72-dbee35a70649

image

Signed-off-by: Lucas ONeil <lucasoneil@gmail.com>

PR #440, opened 2023-02-04
Tenant UI: Credential definition creation and credential offer

loneil:
**Test Data**
To look at this PR can use the following tenants with schemas if you don't want to go through all the setup yourselves.

Issuer
60816750-c2cf-43f4-9089-9fe93da0e637
bb72719e-f729-4dd9-9fcf-bf454d223809

Non-Issuer
3819ccd7-af7d-442f-8f4d-448dfc93ac86
1bd52b6d-5e15-4e35-aec1-a62adf9129e4

**Changes**
Adjust Tenant UI to integrate
- new credential definition creation (for schemas), 
- listing stored cred defs
- credential offer to connections
- general package updates

Only issuer tenants are allowed to do most of these actions or they are disabled in the UI.

Most of this is the original UX designs, I think there can be some revisiting, but should start with this and start getting stakeholder feedback. One thing we'll want to sort out is the "display" name for the credential definitions for usage in lists and dropdowns and things. Have discussed with Sherman and we have some ideas to put in tickets to think about.

Schema creation covered in previous PR here https://github.com/bcgov/traction/pull/438

For an existing Schema, create a credential
image

image
(this can take a bit to complete)

I added a credential def storage page, this hadn't been on designs or there before, but is useful for managing credential defs that you've created. Could possibly just keep it all on the Schemas page too if we want... but useful to have for now.
image

Go to the Issuance page

Offer Credential, select cred and contact from populated lists
image

Can enter fielded attributes from associated schema
image

Or JSON
image

On the Issuance page you can see the state of the credentials you've offered
image

This can work to offer to another agent like a BC Wallet.
After accepting:
image


popkinj (approved):
Lots of progress. Nice work.

PR #441, opened 2023-02-06
Reservation Not Found 428

popkinj:
Fourth and final status of the reservation workflow.
Not found message when there potentially is a valid email and/or id but it is not found in the database.
Screenshot from 2023-02-06 11-48-01

@GurcharanjeetSingh . The error styling is slightly different then the mockup. It's the PrimeVue default message. If you would prefer I align closer to your image that shouldn't be a problem.

### DIFF
--- a/services/tenant-ui/frontend/src/store/reservationStore.ts
+++ b/services/tenant-ui/frontend/src/store/reservationStore.ts
@@ -15,7 +15,7 @@ export const useReservationStore = defineStore('reservation', () => {
       console.log('status', status);
       return true;
     },
-    baseURL: config.value.frontend.apiPath,
+    baseURL: config.value.frontend.tenantProxyPath,
   });
 
   // A different axios instance with a basepath just of the tenant UI backend
--- a/services/tenant-ui/frontend/src/store/reservationStore.ts
+++ b/services/tenant-ui/frontend/src/store/reservationStore.ts
@@ -7,14 +7,7 @@ import { RESERVATION_STATUSES } from '@/helpers/constants';
 
 export const useReservationStore = defineStore('reservation', () => {
   const { config } = storeToRefs(useConfigStore());
-  // A raw api call without using the interceptors from the acapyApiStore
-  // Needed for the open call to reservation at this point
-  // TODO: Allow 404 to pass through to the UI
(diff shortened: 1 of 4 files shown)


loneil (changes requested):
What is this PR trying to do? It's changing the Tenant UI to **NOT** call Traction?
So it's breaking any reservation functionality on the public side because every single status attempt isn't calling the API

services/tenant-ui/frontend/src/store/reservationStore.ts
```diff
@@ -87,10 +83,22 @@ export const useReservationStore = defineStore('reservation', () => {
...
+     * Axios is preconfigured with the "create" method.
+     *
+     * This is a workaround to allow the 404 to be passed to the UI.
+     */
+    await axios(API_PATH.MULTITENANCY_RESERVATION(reservationId), {
+      method: 'GET',
+      // Allow 404 to still approve the promise
+      validateStatus: (stat: number) => stat === 404,
+    })
       .then((res) => {
```
loneil, line 99: This is just 404ing everything since it's trying to request to the frontend app

popkinj: Why do you think that? It's using the same url as the rest of the store.

popkinj: I can do some more testing of different scenarios.

popkinj: Can you give me an example @loneil  where it's not working?

loneil: Try to do any check status on any valid reservation
image

loneil: It's taken out the axios instance that sets it to call the configured API endpoint so that will 404 anything.

popkinj: K. Thanks. I'll have a look.

popkinj: Looks like we're back to square one here. I'll have to look at this being an issue with the proxy server.

PR #443, opened 2023-02-06
Fix CORS errors with error response codes

loneil:
NGINX wasn't returning response codes for error type responses because of CORS, so add the 'always' keyword which I think is the way to overcome that

http://nginx.org/en/docs/http/ngx_http_headers_module.html
```
Adds the specified field to a response header provided that the response code equals 200, 201 (1.3.10), 204, 206, 301, 302, 303, 304, 307 (1.1.16, 1.0.13), or 308 (1.13.0). Parameter value can contain variables.

There could be several add_header directives. These directives are inherited from the previous configuration level if and only if there are no add_header directives defined on the current level.

If the always parameter is specified (1.7.5), the header field will be added regardless of the response code.

```

Reference:
https://serverfault.com/questions/393532/allowing-cross-origin-requests-cors-on-nginx-for-404-responses

### DIFF
--- a/services/tenant-ui/frontend/src/components/LoginForm.vue
+++ b/services/tenant-ui/frontend/src/components/LoginForm.vue
@@ -100,26 +100,28 @@ const handleSubmit = async (isFormValid: boolean) => {
     console.error(err);
     toast.error(`Failure getting token: ${err}`);
   }
-  try {
-    // token is loaded, now go fetch the global data about the tenant
-    await tenantStore.getSelf();
-    console.log(tenant.value);
-    // TODO: once we get response statuses working correctly again can re-configure this
-    // Don't throw errors since not-found and stuff is fine for non-issuers
+  if (token.value) {
     try {
-      // Find out issuer status when logging in
-      await Promise.all([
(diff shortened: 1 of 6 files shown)


popkinj (approved):
Make it so Number 1!

services/tenant-ui/frontend/src/components/LoginForm.vue
```diff
@@ -100,26 +100,28 @@ const handleSubmit = async (isFormValid: boolean) => {
...
+        ]);
+      } catch (err) {
+        console.error(err);
+      }
     } catch (err) {
       console.error(err);
+      toast.error(`Failure getting tenant: ${err}`);
+    } finally {
+      submitted.value = false;
     }
```
loneil, line 124: Cleaning up unrelated double error message display
Will probably adjust this later now that it knows about 404s as well

popkinj:
If this works as intended I'll revert the workaround I did in the store.

popkinj:
If this works as intended I'll revert the workaround I did in the store.

PR #467, opened 2023-02-14
Feature/webhook receiver

usingtechnology:
For local development and illustrative purposes, add a way to catch webhook data for tenants.


popkinj (approved):
Will be a much nicer UI experience to have this working. Nice Job.

esune (approved):

loneil (approved):
Got it working locally end to end (see the webhook results after connections and public did creation)

PR #470, opened 2023-02-15
Typescript types from acapy API

loneil:
Not adding all the interface references in yet (other than a sample/test one), want to get some eyes on the process/layout here first.

Please refer to https://github.com/bcgov/traction/tree/feature/tenantUiCodeCleaning/services/tenant-ui/frontend/src/types to see.

We had tried a openapi-based generator before that had worked with the separate Traction API (which was a nice new FastAPI OpenAPI 3 spec) but that doesn't work with the old Swagger 2.0 Acapy spec. Additionally there seem to be a lot of validation errors in that swagger spec so a lot of other tools didn't seem to be working (going to make issue on acapy repo about this).
Found https://www.npmjs.com/package/swagger-typescript-api which does work.

So going to go with the following strategy

- No need to install any generator as a dependency (it's a develop-time tool, not build or ) so it's not part of any CI/CD process
- If changes happen to the API and updates are needed, just run locally with npx or anything as needed to get a file full of the interfaces and types manually, copy that over to the location of existing file
- Make sure people aren't making any manual changes to the generated file.


esune (approved):
👍🏻

popkinj (approved):
Sounds good to me.

PR #472, opened 2023-02-16
Holder mark when v1 credential revoked

usingtechnology:
Only have v1 credential exchanges in place now, as that is what tenant-ui supports (keep it simple).


usingtechnology:
[issue_and_revoke.txt](https://github.com/bcgov/traction/files/10760022/issue_and_revoke.txt)

rename as python script.

This will illustrate (the holder) getting a list of revoked credentials (`/issue-credential/records?role=holder&state=credential_revoked`)

loneil (approved):

PR #483, opened 2023-02-17
Accept Invitation

loneil:
Will do did exchange in subsequent PR.

This is the invitation URL acceptance for a Traction Tenant. At this point we will only support the connection invitation format, so the URL will need the typical `c_i` param.

image

image

image

### DIFF
--- a/services/tenant-ui/frontend/src/components/contacts/acceptInvitation/AcceptInviteForm.vue
+++ b/services/tenant-ui/frontend/src/components/contacts/acceptInvitation/AcceptInviteForm.vue
@@ -1,5 +1,5 @@
 <template>
-  <form @submit.prevent="handleSubmit(!v$.$invalid)" v-if="urlEntryStep">
+  <form v-if="urlEntryStep" @submit.prevent="handleSubmit(!v$.$invalid)">
     <div class="field">
       <div class="flex justify-content-between">
         <label
@@ -43,14 +43,16 @@
     </div>
   </form>
 
-  <AcceptInviteSubmission v-else />
+  <AcceptInviteSubmission
+    v-else
+    :invitation-string="invitationString"
+    @closed="$emit('closed')"
+  />
 </template>
 
 <script setup lang="ts">
 // Vue
 import { reactive, ref } from 'vue';
-// State
-import { useContactsStore } from '@/store';
 // PrimeVue / Validation
 import Button from 'primevue/button';
 import InputText from 'primevue/inputtext';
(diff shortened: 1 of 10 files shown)


services/tenant-ui/frontend/src/components/contacts/acceptInvitation/AcceptInviteForm.vue
```diff
@@ -96,14 +94,15 @@ const handleSubmit = async (isFormValid: boolean) => {
...
-    emit('closed');
-    toast.info('Invitation Accepted');
+    const inviteParam = paramFromUrlString(
+      formFields.inviteUrl,
+      'c_i'
+    ) as string;
+    if (!inviteParam) {
+      throw Error('Invalid format for invitation URL');
+    }
+    invitationString.value = window.atob(inviteParam);
```
usingtechnology, line 104: not sure if it is worth dealing with at this time but `http:/hey.you?c_i=blah` passes but then blows up because invalid json, kind of buggers up the table load.

services/tenant-ui/frontend/src/components/contacts/acceptInvitation/AcceptInviteSubmission.vue
```diff
@@ -0,0 +1,118 @@
...
+const submitted = ref(false);
+const handleSubmit = async (isFormValid: boolean) => {
+  submitted.value = true;
+
+  if (!isFormValid) {
+    return;
+  }
+
+  try {
+    await contactsStore.recieveInvitation(
```
usingtechnology, line 104: spelling? receiveInvitation...

services/tenant-ui/frontend/src/store/contactsStore.ts
```diff
@@ -85,46 +85,46 @@ export const useContactsStore = defineStore('contacts', () => {
     return invitationData;
   }
 
-  // async function acceptInvitation(inviteUrl: string, alias: string) {
-  //   console.log('> contactsStore.acceptInvitation');
-  //   error.value = null;
-  //   loading.value = true;
+  async function recieveInvitation(invite: string, alias: string) {
```
usingtechnology, line 88: spelling of function should be changed...receiveInvitation

usingtechnology (approved):
couple of comments... for naming of function and optional - what if the c_i parameter doesn't evaluate to json?  not sure if we need to tackle that right now.

other than that - good to go! nice to accept in the app

services/tenant-ui/frontend/src/components/contacts/acceptInvitation/AcceptInviteForm.vue
loneil: Yes, good call.
Validating the JSON using a the form validation to block submission should work (and let the user see what the decoded URL ends up being)
image

services/tenant-ui/frontend/src/components/contacts/acceptInvitation/AcceptInviteSubmission.vue
loneil: argh spelling, fixed

PR #484, opened 2023-02-21
"Manually" load modules as plugins

usingtechnology:
Do not use the plugin registry to load the modules at runtime.

Since the modules aren't true plugins (yet) do not load them via the `plugin_registry`. Load what we need loaded (the setup for each module to hook in listeners and register classes - and routes) directly.  If/when the modules (ie. cred def storage) become separate plugins, they will be configured to load via acapy flags/command line/config files.

The `plugin_registry` alters its internal dictionaries and lists as it loads plugins and routes, so having a plugin that is that is being registered call to register other plugins can cause issues with locks on dictionaries/lists.  What we are doing here should be more stable.


loneil (approved):
Think it makes sense to me. Main innkeeper and storage flows still working on the pr.

Plugins show on the server endpoint
image

PR #488, opened 2023-02-22
Did exchange

loneil:
Additional connection creation method, call DID Exchange Create Request endpoint with a DID and alias.

The person on the **other** end of the exchange, the DID you are entering, they get a connection in their list as well, but they don't have the opportunity to add an alias. So it's like the Mult Use situation again. The "Their Label" field shows the tenant name who initiated this though. So we'll sort all those alias issues soon.

image

image


usingtechnology (approved):
nice work, fired up 2 tenants, got them DIDs and connected via DID Exchange.

PR #492, opened 2023-02-23
Show /plugins list

loneil:
On the About page (Tenant or Innkeeper) in the ACA-Py section fetch the plugins by calling ACA-Py `/plugins`

image

image


loneil:
Tagging @ianco since he's looking at plugin documentation at this moment too.

PR #494, opened 2023-02-23
Plugin: connection update (alias)

### LINKED ISSUE #490: As an ACA-Py API plugin consumer I want to be able to update a connection's alias
This could/should be core aca-py, discussed with Sherman and Ian briefly. For now we could add to a plugin that allows this.

AC:
- Can PUT (or PATCH?) a connections call that will update the alias
- I don't have to supply _all_ connection details to update? Just the alias?

usingtechnology:
Add a `PUT /connection/{connection_id}` endpoint, allow a tenant to update their connection.
*IMPORTANT* only `alias` can be updated. Future fields/properties can be added with community consultation -and if the field should be updated.

Done as a separate plugin so can be used without traction.

resolves #490


loneil (approved):
Great! Can update an alias for an existing connection now. Existing connection behaviour still all good.

image

image

PR #498, opened 2023-02-28
Aca-Py Start Config - maximize env var usage

### LINKED ISSUE #497: Traction Acapy Configuration Model Update
See #493

This is the task to update all Traction Aca-Py startup configuration to use environment variables where possible.

https://github.com/bcgov/trust-over-ip-configurations/blob/main/openshift/templates/issuer-agent/issuer-agent-deploy.yaml

- helm charts
- scripts & docker compose files
- services/acapy (ngrok shell startup for local)

usingtechnology:
Recommendation from @WadeBarnes  to use environment variables where possible, do not use arg file.

WIP.
- reduce scripts/.env-example to used values only
- remove acapy-static-args.yml
- fix up helm charts, values.yaml

Docker/Compose has limitations on duplicate environment variables, so cannot repeat (like `--plugin` and `--plugin-config-value`) so not everything can be an environment variable.

fixes #497


popkinj (approved):
Environment variables for the win!! :tada:

loneil (approved):
Makes sense. Regression from what I can think of to try works out

PR #502, opened 2023-03-04
Credential revocation (and exchange delete)

loneil:
**Revocation**
If an accepted cred is revocable, allow a credential revocation (the X icon, not sure if someone has a better idea, can run it by Guru)
image

When revoking, allow the comment to be filled out if you want
also through some details about the cred revocation (might remove these eventually if they are not useful)
image

Calls /revocation/revoke and then refreshes the table with the updated state.

Tried with BC Wallet and this works (as an aside I'm not sure this revocation/dismiss thing on the card matches the OCA it's trying to go for?)
Screenshot_20230303-173836

When revoking to a cred issued to another Traction tenant this successfully updates the state of the cred exchange record
image
but through the UI right now I'm only showing the HOLDER /credentials list, which doesn't tell you anything about if it's revoked or not?

**Exchange record delete**
Also allow any cred exchange record to be DELETEd as the issuer
image

### DIFF
--- a/services/tenant-ui/frontend/src/assets/tenantuiComponents.scss
+++ b/services/tenant-ui/frontend/src/assets/tenantuiComponents.scss
@@ -34,6 +34,7 @@
   }
 
   /* red */
+  &.credential-revoked,
   &.error,
   &.denied,
   &.rejected,
--- a/services/tenant-ui/frontend/src/components/issuance/IssuedCredentials.vue
+++ b/services/tenant-ui/frontend/src/components/issuance/IssuedCredentials.vue
@@ -12,6 +12,8 @@
     :rows-per-page-options="TABLE_OPT.ROWS_OPTIONS"
     selection-mode="single"
     data-key="credential_exchange_id"
+    sort-field="created_at"
+    :sort-order="-1"
   >
     <template #header>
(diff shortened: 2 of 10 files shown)


esune:
@amanji how is BC Wallet figuring out a credential was revoked in order to display the status?

I am unsure either about which icon to use for revocation, I don't think there is a "standard" UI for it yet - the `X` works for now.

loneil:
> @amanji how is BC Wallet figuring out a credential was revoked in order to display the status?
> 
> I am unsure either about which icon to use for revocation, I don't think there is a "standard" UI for it yet - the `X` works for now.

It must be mapping to the credential exchange record (not just displaying the credential) I'd think. Adding that sort of linking in as part of the ticket I'm working on now for Holder accept/deny/etc. So this PR is just issuer side.

usingtechnology:
> The credential exchange record metadata is updated by the underlying framework when a revocation event is published. We simply use hooks to update the UI when record statuses are updated.

does that make use of a mediator? or is the event pushed to the phone by some other mechanism? just curious as to how events work with mobile.

esune:
@usingtechnology yes, for the mobile wallet a mediator would have to relay the message.

PR #504, opened 2023-03-07
Email 480

popkinj:
One thing left on this branch is to insert the applicant name into the approved/declined templates. This may be the place.... Not sure: https://github.com/bcgov/traction/blob/email-480/services/tenant-ui/frontend/src/components/innkeeper/reservation/ApproveReservation.vue#L27-L59
Looking for feedback from @loneil ☺


popkinj:
I should mention that I've got some test code in here that hardcodes myself as the Innkeeper email recipient. Will change back to the Env Var before merge.

loneil:
@popkinj Yes just pass the name in there as a prop like the email address and supply it to the call

popkinj:
Screenshot from 2023-03-07 10-27-23
Screenshot from 2023-03-07 10-27-30
Screenshot from 2023-03-07 10-27-59
Screenshot from 2023-03-07 12-20-18

GurcharanjeetSingh:
Hi Jamie,
The email body content is good-to-go from the UX end. Just one ask, what about the Subject line of the emails? Will that be as per the UX guidelines in the final production?
Thanks!

popkinj:
> Hi Jamie, The email body content is good-to-go from the UX end. Just one ask, what about the Subject line of the emails? Will that be as per the UX guidelines in the final production? Thanks!

Oh right... I may have missed that  @GurcharanjeetSingh . I'll have another look at the mock-ups.

loneil:
Closing, continued in https://github.com/bcgov/traction/pull/505 (changes made and rebased)

Issue #477, opened 2023-02-17
Spike: Multi-Ledger Write Support


esune:
As a Traction tenant, I want to be able to write to multiple ledgers by submitting the request to the appropriate endorser.

Things to consider:
- is this a "realistic" use-case, or do we expect agents to mostly only target one ledger/network for writes?
- tenant may want to decide which ledgers to request endorsement for (e.g.: only one, all of them. Endorser should be enabled selectively)
- it may be more efficient to just have separate agents on separate Traction instances targeting different ledgers for write purposes

Acceptance Criteria: design document outlining options/implementation steps.

WadeBarnes:
Keep in mind a single tenant would have their resources (Schema(s), Cred-Def(s), etc.) rooted on a single ledger, and the goal would really be to provide the tenants with a choice of which ledger they'd like to be rooted on.

esune:
Closing as done: a tenant will only ever be able to write to one ledger at a time, and the implementation is being handled in #546

PR #506, opened 2023-03-08
R&D only - multiledger investigation/issues

usingtechnology:
this PR is only for sharing code to investigate issues moving from single ledger to multiledger (getting tenants public dids).

see https://hackmd.io/@j2bCRrNkRLC7Nzfg2ypu-g/rJ2gIHSJ2 for more...


esune:
Moving to icebox: this relates to #572 and #477, will be moved along as those issues (and new actionable items logged as result of those) are worked on.

This contains reference material and should not be closed until work is completed, to avoid losing track of it.

PR #528, opened 2023-03-27
Email 514

popkinj:
- Removing a space in the reservation password
- Upgrading the error message to be more useful.

### DIFF
--- a/services/tenant-ui/frontend/src/components/reservation/Status.vue
+++ b/services/tenant-ui/frontend/src/components/reservation/Status.vue
@@ -113,7 +113,6 @@ const spinnerComputed = () => {
   } else {
     icon = 'pi pi-check';
   }
-  console.log('icon', icon);
   return icon;
 };
 
--- a/services/tenant-ui/frontend/src/components/reservation/Status.vue
+++ b/services/tenant-ui/frontend/src/components/reservation/Status.vue
@@ -57,13 +57,15 @@
       }}</small>
     </div>
 
-    <Button type="submit" class="w-full my-2" label="Check Status" />
+    <Button
+      type="submit"
+      class="w-full my-2 check-status-button"
+      label="Check Status"
+      :icon="spinnerComputed()"
+      :disabled="loading"
+    />
   </form>
 
-  <!-- Statuses to check -->
-  <div v-if="loading" class="flex justify-content-center status-spinner">
-    <ProgressSpinner />
-  </div>
   <div>
     <Approved v-if="status === RESERVATION_STATUSES.APPROVED" />
     <CheckedIn v-else-if="status === RESERVATION_STATUSES.CHECKED_IN" />
@@ -81,7 +83,6 @@ import { useRoute } from 'vue-router';
 // PrimeVue/Validation/etc
 import Button from 'primevue/button';
 import InputText from 'primevue/inputtext';
-import ProgressSpinner from 'primevue/progressspinner';
 import { useToast } from 'vue-toastification';
 import { email, required } from '@vuelidate/validators';
 import { useVuelidate } from '@vuelidate/core';
@@ -104,6 +105,17 @@ const toast = useToast();
 // State setup
 const reservationStore = useReservationStore();
 const { loading, status } = storeToRefs(useReservationStore());
+
+const spinnerComputed = () => {
+  let icon = '';
+  if (loading.value) {
+    icon = 'pi pi-spin pi-spinner';
+  } else {
+    icon = 'pi pi-check';
+  }
+  console.log('icon', icon);
+  return icon;
+};
 
 // Login Form and validation
 const formFields = reactive({
@@ -137,8 +149,9 @@ const handleSubmit = async (isFormValid: boolean) => {
   }
 };
 </script>
-<style>
-.status-spinner {
-  transform: scale(0.5);
+<style scoped lang="scss">
+:deep(.check-status-button .p-button-icon) {
+  position: absolute;
+  margin-left: 3rem;
 }
 </style>
--- a/services/tenant-ui/frontend/src/components/reservation/Status.vue
+++ b/services/tenant-ui/frontend/src/components/reservation/Status.vue
@@ -61,7 +61,6 @@
       type="submit"
       class="w-full my-2 check-status-button"
       label="Check Status"
-      icon="pi pi-check"
       :loading="loading"
     />
   </form>
@@ -138,9 +137,3 @@ const handleSubmit = async (isFormValid: boolean) => {
   }
 };
 </script>
-<style scoped lang="scss">
-:deep(.check-status-button .p-button-icon) {
-  position: absolute;
-  margin-left: 3rem;
-}
-</style>
--- a/services/tenant-ui/frontend/src/components/reservation/Status.vue
+++ b/services/tenant-ui/frontend/src/components/reservation/Status.vue
@@ -59,7 +59,7 @@
 
     <Button
       type="submit"
-      class="w-full my-2 check-status-button"
+      class="w-full my-2"
       label="Check Status"
       :loading="loading"
     />
--- a/services/tenant-ui/frontend/src/components/reservation/Status.vue
+++ b/services/tenant-ui/frontend/src/components/reservation/Status.vue
@@ -61,8 +61,8 @@
       type="submit"
       class="w-full my-2 check-status-button"
       label="Check Status"
-      :icon="spinnerComputed()"
-      :disabled="loading"
+      icon="pi pi-check"
+      :loading="loading"
     />
   </form>
 
@@ -105,16 +105,6 @@ const toast = useToast();
 // State setup
 const reservationStore = useReservationStore();
 const { loading, status } = storeToRefs(useReservationStore());
-
-const spinnerComputed = () => {
-  let icon = '';
-  if (loading.value) {
-    icon = 'pi pi-spin pi-spinner';
-  } else {
-    icon = 'pi pi-check';
-  }
-  return icon;
-};
 
 // Login Form and validation
 const formFields = reactive({
--- a/services/tenant-ui/frontend/src/components/reservation/Status.vue
+++ b/services/tenant-ui/frontend/src/components/reservation/Status.vue
@@ -61,10 +61,10 @@
   </form>
 
   <!-- Statuses to check -->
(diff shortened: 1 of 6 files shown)


popkinj:
Screenshot from 2023-03-27 15-32-43

services/tenant-ui/frontend/src/components/reservation/Status.vue
```diff
@@ -61,10 +61,10 @@
   </form>
 
   <!-- Statuses to check -->
-  <div v-if="loading" class="flex justify-content-center">
+  <div v-if="loading" class="flex justify-content-center status-spinner">
     <ProgressSpinner />
   </div>
-  <div v-else>
```
loneil, line 67: We wouldn't want to remove this v-else I think. If loading the status pane below should not be there (albeit each one v-if'd themselves but in case anything else needs to be added)

services/tenant-ui/frontend/src/components/reservation/status/Approved.vue
```diff
@@ -88,7 +94,9 @@ const handleSubmit = async (isFormValid: boolean) => {
       formFields.password
     );
   } catch (error) {
-    toast.error(`Failure: ${error}`);
+    // TODO: Handle expired passwords as well
+    console.error(error);
+    showError.value = true;
```
loneil, line 99: This will show invalid password on any type of error now (network anything, 5xx, Traction down). Think this should be specific to 401 or 403 type errors when the password is actually the issue.
Anyways when handling the expired password later might have to do a string check unless it returns a different error code.

I think it's should be able to just throw a specific error or include the response code in the check-in method

services/tenant-ui/frontend/src/components/reservation/Status.vue
loneil: Yeah need to keep the v-else:

image

usingtechnology:
Also, I haven't been reviewing these, but seeing spelling mistakes... Take care of them in this PR or we can create a new one.

RETRIEVE:  "Are you sure you want to leave this page? You will not be able to retrive these details again" (this is when clicking go back to sign in)

Emails...
RECEIVED  "This email confirms we have recieved your request and a Representative is working on your case. This email confirms we have recieved your request and a Representative is working on your case."

DECISION: "A descision on the request usually takes 2-3 days from the date of this email."

services/tenant-ui/frontend/src/components/reservation/Status.vue
popkinj: > Yeah need to keep the v-else (or do some loading handling on the validate form)
> 
> image

Using the v-else breaks what I'm trying to do. We can maybe restyle it.

loneil: That shouldn't break it at all. What's happening?

popkinj: It may make more sense to move the loading widget into whatever button was pressed. Which I think our UI framework can handle.

services/tenant-ui/frontend/src/components/reservation/status/Approved.vue
popkinj: I was trying to be generic. But we could enhance it with the expired option.
I'm not using docker. How can I adjust the expiry if I'm just linking to dev?

popkinj:
> Also, I haven't been reviewing these, but seeing spelling mistakes... Take care of them in this PR or we can create a new one.
> 
> RETRIEVE: "Are you sure you want to leave this page? You will not be able to retrive these details again" (this is when clicking go back to sign in)
> 
> Emails... RECEIVED "This email confirms we have recieved your request and a Representative is working on your case. This email confirms we have recieved your request and a Representative is working on your case."
> 
> DECISION: "A descision on the request usually takes 2-3 days from the date of this email."

Yeah... May as well fix those here.

services/tenant-ui/frontend/src/components/reservation/Status.vue
loneil: Yeah when validating each of the forms it's probably better to do loaders and disablers on their buttons themselves. Can add that in later.

services/tenant-ui/frontend/src/components/reservation/status/Approved.vue
loneil: Generic works for the user/pw/expiry cases but stuff like traction down or wrong network connection (or any client side error) might not want to trigger a "Your password is incorrect" so I'd throw that only when a 401 or 403 is detected.

Can handle expiry later, if you did want to try it out though but if you don't want to use local docker (would recommend getting that figured out though) then you could set the helm value on a PR to really low (and then undo the commit changing the PR value before merging) with `reservationExpiryMinutes` in the values-pr helm values file. Then config your local tenant UI to the PR.

services/tenant-ui/frontend/src/components/reservation/Status.vue
```diff
@@ -102,6 +103,17 @@ const toast = useToast();
...
+  let icon = '';
+  if (loading.value) {
+    icon = 'pi pi-spin pi-spinner';
+  } else {
+    icon = 'pi pi-check';
+  }
+  console.log('icon', icon);
+  return icon;
+};
+
```
loneil, line 116: Not needed, built into primevue buttons, see how other buttons in the app do it:
image

services/tenant-ui/frontend/src/components/reservation/Status.vue
```diff
@@ -57,14 +57,16 @@
       }}</small>
     </div>
 
-    <Button type="submit" class="w-full my-2" label="Check Status" />
+    <Button
+      type="submit"
+      class="w-full my-2 check-status-button"
+      label="Check Status"
+      :icon="spinnerComputed()"
```
loneil, line 64: See comment below

services/tenant-ui/frontend/src/components/reservation/Status.vue
loneil: Yeah that'll be better then I think (can disable the textboxes conditionally as well if that applies)

popkinj:
Screenshot from 2023-03-28 12-46-16
Screenshot from 2023-03-28 12-47-05

services/tenant-ui/frontend/src/components/reservation/Status.vue
popkinj: Don't know how I missed that. Is it not in the primevue docs?

services/tenant-ui/frontend/src/components/reservation/status/Approved.vue
popkinj: K.. I'm replicating the expired reservation errors now.
I'll show the inline message if there's a 401 and an expiry message... or if there's a 401 and an incorrect password note. Otherwise I'll send the error to Toast.

loneil: That sounds good!

services/tenant-ui/frontend/src/components/reservation/Status.vue
```diff
@@ -57,14 +57,16 @@
       }}</small>
     </div>
 
-    <Button type="submit" class="w-full my-2" label="Check Status" />
+    <Button
+      type="submit"
+      class="w-full my-2 check-status-button"
```
loneil, line 62: check-status-button class remaining

services/tenant-ui/frontend/src/components/reservation/Status.vue
```diff
@@ -134,3 +135,9 @@ const handleSubmit = async (isFormValid: boolean) => {
   }
 };
 </script>
+<style scoped lang="scss">
+:deep(.check-status-button .p-button-icon) {
+  position: absolute;
+  margin-left: 3rem;
+}
+</style>
```
loneil, line 143: Hanging around from previous commit

services/tenant-ui/frontend/src/components/reservation/status/Approved.vue
```diff
@@ -87,10 +105,34 @@ const handleSubmit = async (isFormValid: boolean) => {
...
+    loading.value = false;
   }
 };
 </script>
+<style scoped lang="scss">
+:deep(.validate-button .p-button-icon) {
+  position: absolute;
+  margin-left: 3rem;
+}
+</style>
```
loneil, line 138: Hanging around from previous commit

services/tenant-ui/frontend/src/components/reservation/status/Approved.vue
```diff
@@ -27,7 +27,16 @@
           <small v-if="v$.password.$invalid && submitted" class="p-error">
             {{ v$.password.required.$message }}
           </small>
-          <Button type="submit" label="Validate" class="w-full mt-3" />
+          <Button
+            type="submit"
+            label="Validate"
+            class="w-full mt-3 validate-button"
```
loneil, line 33: validate-button class remaining

services/tenant-ui/frontend/src/components/reservation/status/Approved.vue
```diff
@@ -50,6 +59,7 @@ import { reactive, ref } from 'vue';
 import Button from 'primevue/button';
 import Card from 'primevue/card';
 import Password from 'primevue/password';
+import Message from 'primevue/message';
```
loneil, line 62: Trying to keep these alphabetized where we remember

services/tenant-ui/frontend/src/components/reservation/status/Approved.vue
```diff
@@ -87,10 +105,34 @@ const handleSubmit = async (isFormValid: boolean) => {
...
+      errorMessage.value = 'Reservation has expired.';
+      showError.value = true;
+    } else if (resp.status === 401) {
+      errorMessage.value = 'Incorrect password. Please try again.';
+      showError.value = true;
+    } else {
+      toast.error(resp.data);
+    }
+    console.error('message', resp.data);
+    console.error('status', resp.status);
```
loneil, line 126: Great! Too bad to have to string check for 'expired' but no real other options in terms of API response so all good

loneil (changes requested):
Working great from what I can see! Just minor leftover css from previous icon stuff that is pulling the button icon too far left

image

loneil (approved):
Looking good! 👍

PR #529, opened 2023-03-27
Feature/issue 527 res pwd

### LINKED ISSUE #527: Remove dashes from Reservation passwords
related to #514

Plugin work to simplify the ids and passwords used in reservations, since emails won't have a "copy to clipboard" function and most users (2 out of 2) attempt to select in the email. let's make it easier to double click select a single word. Dashes make that impossible.

usingtechnology:
Remove dashes from reservation password (and reservation id).
These ids end up in emails and having dashes makes it impossible to just double click to select the word. Make it as easy as possible for tenants to complete the reservation flow.

Note: the reservation id is stored as a UUID with dashes, it's only in the REST API that we accept and return without dashes.

Also, touched up the tenant-lob app to use the `/ledger/register-nym` call - should have been done in my last PR, but done now...

Closes #527

### DIFF
--- a/charts/tenant-ui/values-int.yaml
+++ b/charts/tenant-ui/values-int.yaml
@@ -23,5 +23,5 @@ tenant_ui:
     server: apps.smtp.gov.bc.ca
     port: 25
     senderAddress: DoNotReplyTractionINT@gov.bc.ca
-    innkeeperInbox: jason.sherman@gmail.com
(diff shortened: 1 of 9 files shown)


loneil (approved):
Works for me. Get the un-dashed ones to the screen and in the emails, they work to check in to create a new tenant fine.

PR #530, opened 2023-03-28
Check Status return route for reservation email. Update libs

loneil:
While we're doing other reservation improvements, add a return to /check-status so the reservation-maker has a link back from their email to the check status section (this was in the original email specification)

image

image

Also npm updates

Also regen acapy interfaces to pick up changes


esune:
Good idea about pre-populating the fields from the link. Should we be looking into obfuscating the link though, rather than have plaintext query params?

This looks good! 👍🏻

esune (approved):

loneil:
That is a great idea. It could just be a link back with query params yeah (url encoded)
so to `/check-status?email=lucas.oneil%40gov.bc.ca&id=1221d4cd31874480baed232b001c9893`

the email could be base64 encoded or something if that's necessary @esune

esune:
> That is a great idea @usingtechnology . It could just be a link back with query params yeah (url encoded) so to `/check-status?email=lucas.oneil%40gov.bc.ca&id=1221d4cd31874480baed232b001c9893`
> 
> the email could be base64 encoded or something if that's necessary @esune

I was thinking a fully-obfuscated query param (like `myparam=<base64 string here>`) containing everything, or something to limit intercepting the data - I don't think the status check function opens up any real attack vector otherwise. Can be part of another issue as well depending on complexity.

loneil:
> > That is a great idea @usingtechnology . It could just be a link back with query params yeah (url encoded) so to `/check-status?email=lucas.oneil%40gov.bc.ca&id=1221d4cd31874480baed232b001c9893`
> > the email could be base64 encoded or something if that's necessary @esune
> 
> I was thinking a fully-obfuscated query param (like `myparam=<base64 string here>`) containing everything, or something to limit intercepting the data - I don't think the status check function opens up any real attack vector otherwise. Can be part of another issue as well depending on complexity.

Added in auto-populate ability. Just normal query params right now (easy to handle with vue router)
I just have them as hyperlink, not the actual link text in the email for clarity/readability. Could discuss if that's the right call but makes sense to me.
Should see more user feedback once this and Jamie and Sherman's changes are in, think together they'll eliminate a lot of confusions

image

PR #538, opened 2023-04-04
create and list OCA

loneil:
Layer for accessing OCA endpoints in traction plugin
https://traction-tenant-proxy-dev.apps.silver.devops.gov.bc.ca/api/doc#/oca

New OCA page (must be issuer since it uses creds you have in the system to choose from, could potentially remove that need but not sure if that makes sense)

image

image

image

### DIFF
--- a/services/tenant-ui/frontend/package.json
+++ b/services/tenant-ui/frontend/package.json
@@ -15,6 +15,7 @@
     "@vuelidate/validators": "^2.0.1",
     "axios": "^1.3.4",
     "date-fns": "^2.29.3",
+    "json-editor-vue": "^0.10.5",
     "jwt-decode": "^3.1.2",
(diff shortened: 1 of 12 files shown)


services/tenant-ui/frontend/package.json
```diff
@@ -15,6 +15,7 @@
     "@vuelidate/validators": "^2.0.1",
     "axios": "^1.3.4",
     "date-fns": "^2.29.3",
+    "json-editor-vue": "^0.10.5",
```
loneil, line 18: Simple vue wrapper for the existing `vanilla-jsoneditor` package we use that saves having to init it in script setup and lets us use a vue component easily

esune (approved):
Looks good to me 👍🏻 
Adding @amanji as reviewer in case he has input and for general awareness as well.

PR #549, opened 2023-04-18
Bundle loading and initial OCA view (simple card only) in Holder credentials

loneil:
Add OCA loading (from URL) from Traction call (see note below) for a credential-OCA association.
Add simple card view initial implementation from loaded OCA.

This PR has a tenant with a couple accepted creds with OCA implementations
c2ca5a85-d453-4f48-b85b-7501eaadf833
89051761-79f9-4f82-9ce2-882a7bc4fc91

For the layout and composition of the OCA cards, this should be considered as POC implementation, we don't want to go too far down finalizing a Vue-specific style and layout structure until we consider how we can make a more generalized library so this doesn't have to be repeated in every app for every language. Discussed with @esune and @amanji .

The code for parsing the bundle into TS objects has been mostly copied over from @amanji 's great work in https://github.com/bcgov/aries-oca-bundles/tree/main/OCAExplorer
So that's one of the places we will not want to touch much as part of this app as it can be extracted to an NPM package.

image
(student card in this example has 404 in linked image from OCA)

Notes/future tickets to track
- There seems to be an issue with the Traction plugin OCA endpoint for loading associations made by the credential owner, it's supposed to work with public calls but that isn't returning at this point... so for this example we're self-connecting and issuing. https://github.com/bcgov/traction/issues/574
- Need to adapt language dropdown to pass into OCA card (and standardize language codes) https://github.com/bcgov/traction/issues/575


esune:
FYI @loneil I linked the issue I *think* is appropriate for this

swcurran:
I think it is really bad practice to call an PR “WIP” since there is a flag for that…unless of course this is about adding a “WIP” feature to Traction.  :-).

services/tenant-ui/frontend/src/store/contactsStore.ts
```diff
@@ -40,6 +40,16 @@ export const useContactsStore = defineStore('contacts', () => {
     contacts.value.filter((c) => c.state === CONNECTION_STATUSES.INVITATION)
   );
 
+  const findConnectionName = computed(() => {
+    return (connectionId: string) => {
+      // Find the connection alias for an ID
+      const connection = contacts.value?.find((c: any) => {
+        return c.connection_id === connectionId;
+      });
+      return connection && connection.alias ? connection.alias : '...';
```
esune, line 49: Is this returning the `...` for the UI? Would `undefined` be a better, more explicit option?

loneil: Returning ... for loading state since the connection loading could be going while the element wanting to display it could be on there.
Could refactor for different logic for not found vs contacts loading
https://github.com/bcgov/traction/issues/576

esune (approved):

Issue #465, opened 2023-02-10
TenantUI: Set all strings to be i18n resource strings


loneil:
Best-effort for no hardcoded strings in the TenantUI

AC:
- Any screen text (public, tenant, and Innkeeper) is replaced with the vue i18n functions
- All strings are contained in the example translation json files
- Best-effort for this, focus on templates, if any javascript stuff is blocking or not working can raise a ticket and circle back to later

PR #568, opened 2023-05-04
Convert all raw strings to i18n translated

amanji:
- Added devcontainer config
- Added vue-i18n eslint plugin and automation scripts
- Converted hard-coded strings to i18n
- Updated package.json npm run commands
- Fixed build errors

_Note: this will require another pass through to ensure all i18n translations are in place_

Fixes: #465


loneil:
Seeing some odd behaviour I haven't before where some local strings don't appear to be resolving but only intermittently.
Observed on PR deployment but screenshot below is local

image

Not sure if there's any change from using that global `$t` rather than the composable that was there before?
The **first** one in the screenshot is legitimately missing (should be `common.walletId`) whereas the rest are clearly working on the left but not the right...
Maybe having one missing is short-circuiting, or there's some race condition with loading the files now or something

loneil (approved):
Can't reproduce issue from screenshot yet any more so can leave for now I think and fix forward any missing keys and raise a ticket if we see that again

PR #578, opened 2023-05-10
Allow urls without tokens

popkinj:
This is now working as intended now.
Screenshot from 2023-05-10 10-28-00


loneil:
Seems to let you add blanks now (which doesn't happen in main branch). We should not save the item if the whole thing is empty string.

image
(ignore i18n strings in this screenshot, will pr a fix to some of the missing ones in a bit)

loneil:
Think you just need an `if` statement somewhere in the foreach loop on line 225

loneil (approved):

loneil:
@popkinj please rebase PR to update latest changes

PR #580, opened 2023-05-15
Placeholder UI for to-come ACA-Py per-tenant settings

loneil:
Do not merge in current state, PR is showing Tenant UI part, will comment out before merge.

Tenant on this PR env for testing
27bd4cf6-2b0b-4ac1-bc0a-6e9c49839d04
d7528d8d-f45b-42db-84d0-6b158b4e81ae

This is the Tenant UI frontend settings for the per-tenant ACA-Py settings currently being developed. Implementation to come (in ) is to be an "extraSettings" property in the POST body for 
https://traction-tenant-proxy-dev.apps.silver.devops.gov.bc.ca/api/doc#/multitenancy/put_multitenancy_wallet__wallet_id_
with each of these fields in it.

This PR adds the Tenant UI part in advance.
We can just comment it all out while waiting for the ACA-Py part to be developed and released and then add it back in once that stuff is there.

Simply a bunch of new flags and fields on the Settings page. Exact layout and wording can be adjusted later (could collapse under a collapsible section or something as well)
Settings are just in alphabetical order but could find out later with user feedback if some are more top-line or important than others. Grouped the "auto" flags together as well.

Hover help texts on the '?' come from the field descriptions here
https://github.com/hyperledger/aries-cloudagent-python/blob/main/aries_cloudagent/config/argparse.py

image

Added an alert while this is temporary to show the fields that will be posted
image


esune:
Looks good to me. Agree on eventually arranging/grouping fields together based on their function. Could also start as mentioned by putting them under a collapsible section to avoid the page from being super long from the get go.

PR #586, opened 2023-05-17
Feature/taa accept

loneil:


services/tenant-ui/frontend/src/helpers/index.ts
```diff
@@ -24,14 +24,24 @@ export function formatDate(value: string) {
...
 }
 
+/**
+ * @function formatUnixDate
+ * Converts a unix time number to an 'MMMM D yyyy' formatted string
+ * @param {String} value A unix timestamp number
+ * @returns {String} A string representation of `value`
+ */
+export function formatUnixDate(value: number) {
+  return format(fromUnixTime(value), 'MMMM d yyyy');
```
esune, line 42: Could  we put the date format in the i18n file, so it can be localized?
Same for other, pre-existing values.

esune:
Being nitpicky here: could we display the mechanism text as well as the key in the approval method selector? Maybe a radio button list or something like that. I am just thinking the human-readable descriptive text would be better - the drawback is that it cannot go through i18n

services/tenant-ui/frontend/src/helpers/index.ts
loneil: can use date-fns/locale in these I think, adding https://github.com/bcgov/traction/issues/593

loneil:
> Being nitpicky here: could we display the mechanism text as well as the key in the approval method selector? Maybe a radio button list or something like that. I am just thinking the human-readable descriptive text would be better - the drawback is that it cannot go through i18n

@esune added radios with text
image

esune (approved):

Issue #587, opened 2023-05-18
Update ngrok images


esune:
The current `wernight/ngrok` images are outdated, and need to be updated to use the official `ngrok/ngrok` images.
When starting-up the project, the `ngrok` containers print this error message:
```log
t=2023-05-18T11:52:45+0000 lvl=info msg="no configuration paths supplied"
t=2023-05-18T11:52:45+0000 lvl=info msg="using configuration at default config path" path=/home/ngrok/.ngrok2/ngrok.yml
t=2023-05-18T11:52:45+0000 lvl=info msg="open config file" path=/home/ngrok/.ngrok2/ngrok.yml err=nil
t=2023-05-18T11:52:45+0000 lvl=info msg="starting web service" obj=web addr=0.0.0.0:4040
t=2023-05-18T11:52:45+0000 lvl=eror msg="failed to auth" obj=tunnels.session err="Your ngrok agent version \"2\" is no longer supported. Only the most recent version of the ngrok agent is supported without an account.\nUpdate to a newer version with `ngrok update` or by downloading from https://ngrok.com/download.\nSign up for an account to avoid forced version upgrades: https://ngrok.com/signup.\r\n\r\nERR_NGROK_120\r\n"
t=2023-05-18T11:52:45+0000 lvl=eror msg="failed to reconnect session" obj=csess id=d7f4af08a79c err="Your ngrok agent version \"2\" is no longer supported. Only the most recent version of the ngrok agent is supported without an account.\nUpdate to a newer version with `ngrok update` or by downloading from https://ngrok.com/download.\nSign up for an account to avoid forced version upgrades: https://ngrok.com/signup.\r\n\r\nERR_NGROK_120\r\n"      
t=2023-05-18T11:52:45+0000 lvl=eror msg="session closing" obj=tunnels.session err="Your ngrok agent version \"2\" is no longer supported. Only the most recent version of the ngrok agent is supported without an account.\nUpdate to a newer version with `ngrok update` or by downloading from https://ngrok.com/download.\nSign up for an account to avoid forced version upgrades: https://ngrok.com/signup.\r\n\r\nERR_NGROK_120\r\n"
t=2023-05-18T11:52:45+0000 lvl=info msg="received stop request" obj=app stopReq="{err:0xc0005500d0 restart:false}"
[...]

Issue #583, opened 2023-05-15
Use a Traction instance for the May 31 2023 AnonCreds Workshop


swcurran:
I'm (delighted to be) on the hook to host an AnonCreds workshop with Hyperledger on May 31.  Last I heard we had over 200 people signed up.  I can create the slides and materials easily, but it is called a workshop, and I'm working with Rodolfo Mirando of Roots.ID on that part.  We need a way to have users try out creating a credential, issuing it and verifying it, where more time is spent on the AnonCreds part, and less on the Aries part.

Akiff, Hannah, Aaron and I talked at OSS and think that Traction could be a great way to do this.  We would have each participant requisition and receive a Traction agent. They use that to create a schema or two, connect to a holder and issue a credential (or two) and they do a presentation request with multiple credentials.

To do that we need some effort on the Enterprise Apps team (yes, these will go into issues if we want to go forward):

- deploy (if needed) an instance of Traction that can be used.  Or use an existing one.
- make sure we can support the load -- we'll have as many participants as possible get their instance before the workshop.
- Use BC Wallet as the holder — display QR codes for the interactions.
   - Might also enable having the agent to connect to itself as the holder -- so that participants only need one agent.  Evidently that doesn't work today for some reason.
- Construct a script for them to follow -- with space for them to do their own playing with the schema and the presentation request.

Confirming we can use the BC Wallet — that it is available globally,

More details to follow.

esune:
We should be able to use https://tenant-ui-dts-dev.apps.silver.devops.gov.bc.ca as instance - c.c.: @WadeBarnes 
For the load testing, can we use @usingtechnology script/load tester to figure out the provisioning part? The "genera" usage should be fine, I think, maybe we just tweak and increase the resources a bit.

esune:
@swcurran instance has been deployed ( see https://github.com/bcgov/DITP-DevOps/issues/74 ) and credentials to innkeeper provided separately.

Do you need a hand building/improving the script to follow, or are we good as long as the instance is up and running?

PR #588, opened 2023-05-18
Feature/auto approve reservation

### LINKED ISSUE #584: Fully-Automatic Tenant Registration Flow
For situations such as what is described [here](https://github.com/bcgov/traction/issues/583), it would be nice to have a fully-automatic tenant approval option that will not require the innkeeper to manually approve one by one each registration request.

Acceptance Criteria:
Add an environment variable to be set at deployment time that, if present and set to `true`, automatically approves the tenant registration request.

The tenant will still be responsible for completing the `check-in` step and obtaining the credentials to access/control their agent.

esune:
Add setting to auto-approve tenant reservations - useful for situations such as described in #583.
UI was updated to show password as well - it could use some extra tweaks, potentially.

Resolves #584, #587


services/tenant-ui/frontend/src/components/reservation/ReservationConfirmation.vue
```diff
@@ -9,6 +9,9 @@
         {{ $t('reservations.submitted') }} <br />
         {{ $t('reservations.emailSentTo', [email]) }}
       </p>
+      <p class="text-center" v-if="pwd">
```
loneil, line 12: Can hit an `npm run lintfix` in `frontend` to auto-fix these lint errors

scripts/plugin-config.yml
```diff
@@ -15,4 +15,6 @@ traction_innkeeper:
     wallet_key: change-me
     print_key: true
     print_token: true
-  reservation.expiry_minutes: 2880
+  reservation:
+    auto_approve: true
```
loneil, line 19: Might need to be added through to the helm value yamls and stuff? Same spots as the "expiry_minutes" in a few places?

loneil:
Pulled locally and ran with local Docker and works all good.
We could short-cut this part on the UI for sure since the password returned could be used to just get the wallet automatically so the user could probably just do it with a 1 step thing. Could make a backlog ticket for those changes or something.

image

services/tenant-ui/frontend/src/components/reservation/ReservationConfirmation.vue
esune: Thanks for the tip, run a lint and fixed them 🙂

scripts/plugin-config.yml
esune: Good catch, I completely forgot. Added the setting to the base `values.yaml` - environment specific values can override the setting, that will otherwise be set to `false`.

esune:
> Could make a backlog ticket for those changes or something.
> 

Logged https://github.com/bcgov/traction/issues/592 to track this. I think it is fairly low priority though, given the edge use-case and the fact that there is a way to deal with it.

esune:
Errors appear to be related to the PR being opened from a fork, rather than from a branch in the repository. Merging as the changes are required to continue working on https://github.com/bcgov/DITP-DevOps/issues/74

PR #589, opened 2023-05-18
InnKeeperWalletConfig update and access mechanism

### LINKED ISSUE #540: Per-Tenant settings storage in plugin
The traction plug-in needs to store settings that are specific to each tenant, such as:
- whether they can connect to the endorser or not (#393)
- whether their did can be public or not (#394)
- other

Touch base with @usingtechnology to review what might already be available.

shaangill025:
- resolve #540 
- Add attributes in `InnKeeperWalletConfig` regarding connection to endorser and registering public DID
- Access to these attributes when connecting to endorser [inside plugin] and registering public DID [tenant-ui]
- These attributes are not being currently used. But it will be with multi-ledger support.

### DIFF
--- a/plugins/traction_innkeeper/traction_innkeeper/v1_0/innkeeper/tenant_manager.py
+++ b/plugins/traction_innkeeper/traction_innkeeper/v1_0/innkeeper/tenant_manager.py
@@ -7,10 +7,13 @@ from aries_cloudagent.core.error import BaseError
 from aries_cloudagent.core.profile import Profile
 from aries_cloudagent.messaging.models.base import BaseModelError
 from aries_cloudagent.multitenant.base import BaseMultitenantManager
+from aries_cloudagent.ledger.multiple_ledger.base_manager import (
+    BaseMultipleLedgerManager,
+)
 from aries_cloudagent.storage.error import StorageError, StorageNotFoundError
 from aries_cloudagent.wallet.models.wallet_record import WalletRecord
 
-from .config import TractionInnkeeperConfig
+from .config import TractionInnkeeperConfig, InnkeeperWalletConfig
 from .models import TenantRecord, ReservationRecord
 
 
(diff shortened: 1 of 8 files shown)


plugins/traction_innkeeper/traction_innkeeper/v1_0/innkeeper/tenant_manager.py
```diff
@@ -124,10 +127,19 @@ async def create_tenant(self, wallet_id: str, tenant_id: str = None):
...
         wallet_key = config.wallet_key
+        # multi_ledger_manager = self._profile.inject(BaseMultipleLedgerManager)
+        self._profile.settings.set_value(
+            "tenant.endorser_config",
+            config.connect_to_endorser,
+        )
+        self._profile.settings.set_value(
+            "tenant.public_did_config",
+            config.create_public_did,
+        )
```
esune, line 142: Can these be specified per-tenant? If it is a setting at the InnKeeper wallet level, how do we specify for each tenant which one can have a public DID and/or become an issuer?

loneil: Yeah the InnkeeperWallet naming here confused me as well at first :)

loneil:
Would each **tennant** be calling `/innkeeper/wallet-config/${id}` to get their settings?

Is it necessary due to the plugin design? Tenants call things under `/tenant` (see https://traction-tenant-proxy-dev.apps.silver.devops.gov.bc.ca/api/doc#/traction-tenant) to get info about their tenant in the Innkeeper plugin so that could be good if this was also under there.

Although that is "tenant self-administration", whereas this is administered by the innkeeper....

Not sure if other opinions on that or limitations
@esune @shaangill025

shaangill025:
@loneil There is no technical compulsion to put the endpoint under `innkeeper` rather than `traction-tenant`, I agree that it should be under `traction-tenant`. I had initially kept under `innkeeper` as I got confused by `InnkeeperWalletConfig` label.

esune (approved):
Looks good to me now - the endpoints look much more straightforward with the latest update, thanks!

loneil (approved):
Looks good to me 😄

PR #556, opened 2023-04-25
Upgrade PostgreSQL to v14.5, Update Helm Chart files

i5okie:
Addresses https://github.com/bcgov/traction/issues/539

- Update `docker-compose.yml` to use `postgres:14.5` image
- Update Helm chart files with learnings from https://github.com/bcgov/traction/issues/539#issuecomment-1518046052
- Update Helm chart files as required to use updated Bitnami PostgreSQL chart v11.9.13 (PostgreSQL v14.5)

`traction-dts` deployment has been upgraded.

## How to upgrade
- Scale down app deployments to 0
- Perform database backup of every database within postgres statefulset (askar-wallet, mywallet/default)
- Update appropriate values files (use `values.yaml` and `values-e79518-dev.yaml` for reference. Make sure to update postgresql values.
- Make sure `wallet-key` and `wallet-name` values are set the same as your previous deployment.
- Delete postgres statefulset and its PVC
- Perform Helm upgrade using your values files (app deployments should stay scaled down)
- Restore databases to the new postgres statefulset
- Scale app deployments back up

#### Note
If `wallet-name` was not previously set, it is likely to be `default`

PR #594, opened 2023-05-19
Fix for tenant UI CI/CD on PRs (Add back overrides)

loneil:
The Tenant UI still appears to need the override altering for fullNameOverride and tractionNameOverride.
Not sure if it was intended to have them taken out in https://github.com/bcgov/traction/pull/556 ?
Seems to be ok in the Traction part after those were removed so maybe there's a different mechanism intended for PR number adjustment.

But we'll need this back in for now at least


esune (approved):
Let's put this back at least for now. @i5okie take note as you'll be looking at cleaning up charts in general soon-ish 🙂

PR #605, opened 2023-05-24
feat: vitest testing framework

amanji:
- feat: add vitest and coverage package
- feat: adds a basic test, remove coverage output
- chore: remove yarn lock
- feat: add snapshots


services/tenant-ui/frontend/test/__snapshots__/basic.test.ts.snap
esune, line 1: Are these linting errors?

amanji: No they are reference test snapshots. They get included such that if there are any subsequent changes to the component the changes must be accounted for, typically to prevent regressions. 

https://vitest.dev/guide/snapshot.html

esune: I meant the code highlighted in dark (red?), not the files themselves.

amanji: Ah ok I figured that's what you meant. I haven't seen any linter failures from them. But I can also see if snapshots can be output in a different format.

esune: Nvm, I thought this looked similar to some linter errors highlighted in one of my previous PRs, but I think it is not the same (I actually got a GHA badge stating what the error was).

esune (approved):
👍🏻 
I'd wait for input from the other reviewers before merge though

loneil (approved):
Seem to be able to `npm run test` locally so think it's looking good!

Get one warning at the beginning but maybe something to resolve later
```
[Vue warn]: Failed to resolve component: router-view
If this is a native custom element, make sure to exclude it from component resolution via compilerOptions.isCustomElement.
  at <App ref="VTU_COMPONENT" >
  at <VTUROOT>
  ```

Not too knowledgeable on those snapshot files so yeah as long as it's passing lint the highlighting is ok for now. I'll familiarize more with vitetest soon as well.

PR #606, opened 2023-05-26
Skip reservation password entry for auto-approve and check in automatically

loneil:
**NOTE: In this PR, auto-approve is on, can undo flag setting before merge**

Bypass reservation ID and PW if the reservation PW comes back to the Tenant UI (IE, auto-approve is on).
Just call check-in with the returned details. Some view changing to handle this flow. Tested it out with auto on and auto off and working fine.
Skips sending the email with the reservation ID and PW since the user has no concept of those in this case. The Wallet ID and Key are handled all the same as before.

image
--
image
--
image

### DIFF
--- a/charts/traction/values-pr.yaml
+++ b/charts/traction/values-pr.yaml
@@ -21,6 +21,7 @@ acapy:
       printKey: true
       printToken: true
       reservationExpiryMinutes: 2880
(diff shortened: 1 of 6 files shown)


esune (approved):
👍🏻

Issue #455, opened 2023-02-10
Run tests in GH actions pipeline


loneil:
AC:
- When committing a PR unit tests are run against the Tenant UI backend and the tenant UI frontend as 2 separate jobs.
- If a fail occurs the PR is blocked with the ❌

PR #615, opened 2023-05-26
feat: Tenant UI GHA test workflow

amanji:
Workflow to run test suites for Tenant UI when PR or push to `develop` branch.

* Test suites are separated into `frontend` and `backend`

Fixes: #455

### DIFF
--- a/.github/workflows/test_tenant_ui.yml
+++ b/.github/workflows/test_tenant_ui.yml
@@ -18,7 +18,7 @@ jobs:
 
     strategy:
       matrix:
-        node-version: [16.x, 18.x, 20.x]
+        node-version: [18.x, 20.x]
         # See supported Node.js release schedule at https://nodejs.org/en/about/releases/
 
     steps:
@@ -41,7 +41,7 @@ jobs:
 
     strategy:
       matrix:
-        node-version: [16.x, 18.x, 20.x]
+        node-version: [18.x, 20.x]
         # See supported Node.js release schedule at https://nodejs.org/en/about/releases/
 
     steps:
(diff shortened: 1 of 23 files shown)


.github/workflows/test_tenant_ui.yml
```diff
@@ -0,0 +1,57 @@
...
+jobs:
+  backend:
+    runs-on: ubuntu-latest
+    defaults:
+      run:
+        working-directory: ./services/tenant-ui
+
+    strategy:
+      matrix:
+        node-version: [16.x, 18.x, 20.x]
```
esune, line 21: Unless we still depend on `16.x` for some reason I would remove it, it is EOL anyway in September of 2023.

services/tenant-ui/jest.config.ts
```diff
@@ -162,9 +162,7 @@ export default {
   // ],
 
   // An array of regexp pattern strings that are matched against all test paths, matched tests are skipped
-  // testPathIgnorePatterns: [
-  //   "/node_modules/"
-  // ],
+  testPathIgnorePatterns: ["/frontend/"],
```
esune, line 165: Are we ACTUALLY supposed to skip tests in frontend? 🤔

esune (changes requested):

services/tenant-ui/jest.config.ts
amanji: There are two testing frameworks in place. The "backend" was already using Jest while the frontend is using Vitest based on the suggestion by @loneil. The GHA run tests for both separately. If I don't ignore the frontend module when running the Jest tests, those tests will fail because Vitest is not installed or configured. We could align and use the same testing library for both.

This also brings up a larger question I have about why the tenant ui backend has the frontend nested within it? Why not have the backend and frontend folders setup as siblings?

loneil: Oh yeah that's just the pattern I've always used to do, probably no specific good reason to so we could reconfigure I'm sure at some point easily enough

services/tenant-ui/Dockerfile
```diff
@@ -1,13 +1,16 @@
 FROM node:16-alpine
```
esune, line 1: Might want to update the image version since we excluded it from the tests 😉

esune (approved):

PR #630, opened 2023-05-31
fix: add Vite plugin to load i18n messages

amanji:
This PR fixes an issue where untranslated keys were intermittently displaying in the UI. This was caused by language files not loading properly. 

The `vue-i18n` Vite plugin solves this  issue.


loneil (approved):
Awesome stuff! Can't seem to make the bad thing happen on deployed PR, pulled locally and running in Vite dev and build it's looking clean too! 👍

Issue #478, opened 2023-02-17
Add filter function to tabular data in Traction Tenant UI screens


krobinsonca:
As a Tenant UI user, I want to filter tabular data, so I can find records I'm looking for

See for example: https://primevue.org/datatable

AC:
- All tables' columns have filters for appropriate columns
- Best effort at not having a ton of boilerplate duplication
- Matches Guru's mockups, or if Primevue deviates too much, discuss design compromise with Guru

hiteshgh:
final testing WIP

loneil:
@hiteshgh need a pull request to review before moving code tickets into review

Gavinok:
> @hiteshgh need a pull request to review before moving code tickets into review

Sorry Lucas that was my mistake. I thought since it says review/QA that it would be moved there during testing.

Issue #452, opened 2023-02-10
Tenant UI: Reduce code duplication


loneil:
Fairly vague to start here... general DRY stuff

ideas
- tables in components
- store API actions, (try, catch, finally, loading, error, return)

AC:
- Things feel good? 5 points of effort spent? I don't know

hiteshgh:
Lucas and Akiff to divide and conquer this ask and we can discuss in next planning meeting

PR #644, opened 2023-06-05
Add filters to table

Gavinok:
Added filters to datatables as described in #478 as of now all filters are done via text input. Theoretically these filters could be changed to act as switches, a calendar view for dates etc. due to time constraints and the overhaul needed for that I think we should hold off until #452 is completed.

 Each filter is given a dedicated button to display as shown below
image

When clicked a text input prompt is given where the user can filter the selected contents
image

Alternatively there is a row view rather than the menu option. However, the menu option provides better support for future work adding new filter types as mentioned above.


loneil:
@Gavinok we'll have to have this one be changed to be from a branch on the Traction repo rather than pull requests from a fork.
Added you as a maintainer to the repo.

Gavinok:
replaced with #649

PR #649, opened 2023-06-06
Add filters to table

Gavinok:
Added filters to datatables as described in #478 as of now all filters are done via text input. Theoretically these filters could be changed to act as switches, a calendar view for dates etc. due to time constraints and the overhaul needed for that I think we should hold off until #452 is completed.

 Each filter is given a dedicated button to display as shown below
image

When clicked a text input prompt is given where the user can filter the selected contents
image

Alternatively there is a row view rather than the menu option. However, the menu option provides better support for future work adding new filter types as mentioned above.


services/tenant-ui/frontend/src/store/utils/fetchList.ts
```diff
@@ -2,9 +2,9 @@ import { useAcapyApi } from '../acapyApi';
 import { AxiosRequestConfig } from 'axios';
 import { Ref } from 'vue';
 
-export async function fetchList(
+export async function fetchList<T>(
   url: string,
-  list: Ref<any>,
+  list: Ref<T[]>,
```
loneil, line 7: Yes, thank you! Been meaning to add this at some point for better TS typing.

loneil:
Sample tenant for PR env

0ed8591b-a9d7-46ca-8bf0-d304ecc5f025
87114f99-80be-42ab-80f0-2091a95a96f7

loneil (changes requested):
Don't seem to be able to click on other items than "Contains" in the filter popup

image

And trying spits some errors in the console (these from my local to see Vue warnings as well)

image


Probably ok to just have Contains without the other options if they can't work, but then we should hide the dropdown if that's available through the table settings, or if no other option use a global CSS to hide it.

loneil:
@Gavinok can we also add to the Innkeeper (reservations and history) tables as well? Thanks.
See https://pr-649-tenant-ui-dev.apps.silver.devops.gov.bc.ca/innkeeper (can find login credentials for lower environments in default.json if you don't know them already)

Gavinok:
@loneil I have removed the option to select the filtering mode and support for filtering in the innkeeper. Should resolve those requested changes

loneil (approved):
Looking good, should be fine with just the `contains` as the only option 👍

loneil (approved):

PR #660, opened 2023-06-07
BUG: Fix cred table identifiers

loneil:
I didn't notice that some of the tweaking of table identifiers for the Filtering PR broke the ID on the credential exchange table, so just fixing that quickly here.

Problem in Innkeeper identifier as well, fixing that.


loneil:
Merging this as it's a quick fix and actively causing issues in dev env.

PR #661, opened 2023-06-08
Add ngrok authtoken setting

esune:
Add optional ngrok authtoken setting

### DIFF
--- a/scripts/.env-example
+++ b/scripts/.env-example
@@ -1,4 +1,5 @@
 TRACTION_ENV=local
+# NGROK_AUTHTOKEN=<your token goes here>
 
 # ------------------------------------------------------------
 # Traction configuration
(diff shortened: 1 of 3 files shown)


amanji (approved):

scripts/.env-example
```diff
@@ -1,4 +1,5 @@
 TRACTION_ENV=local
+# NGROK_AUTHTOKEN=<your token goes here>
```
swcurran, line 2: Quick question — I don’t know the answer. Is it best practices to put the token here in the file, or better to set it in your environment?  Thanks

esune: This is the sample file provided with the repo, you would set it in the `.env` file that is copied (and gitignored) and does not get checked back into the repo.

You could, alternatively, do something like `NGROK_AUTHTOKEN=... docker compose up` - I think, but this would need to be remembered every time you restart the services in docker. It might be okay if you plan on having long-running instances though.

swcurran: Thanks — the latter was what I was asking about.  Sounds good!

Issue #611, opened 2023-05-26
Add Traction token details to the Developer page


loneil:
As a Tenant UI developer user, I want to see the current token on the Developer page, so that I can easily copy it to use in Swagger/Postman etc calls.
--

AC:
- The Developer page (where the config is shown) spits out the current token
- It has a copy to clipboard button (prepend "Bearer " in front too)
- Show or toggle to the decoded JWT object as well

PR #669, opened 2023-06-14
Add JWT to Developer view in tenant UI

Gavinok:
Added support for viewing encoded and decoded JWTs as well as quickly copying them to clipboard resolving #611.

image

When the copy button is clicked a notification is displayed to inform them

image


services/tenant-ui/frontend/src/components/profile/Developer.vue
```diff
@@ -2,14 +2,67 @@
   <!-- eslint-disable-next-line @intlify/vue-i18n/no-raw-text -->
   <h2>{{ $t('profile.config') }}</h2>
   <vue-json-pretty :data="config" />
+  <h2>{{ $t('profile.token') }}</h2>
```
amanji, line 5: Good practice moving forward is to maybe encapsulate these into their own components

amanji (approved):

amanji:
You'll need to update your branch before it can be merged

amanji (dismissed):

services/tenant-ui/frontend/src/components/profile/JWT.vue
```diff
@@ -0,0 +1,60 @@
+<template>
+  <h2>{{ $t('profile.token') }}</h2>
+  <Accordion :multiple="true">
+    <AccordionTab header="Encoded JWT">
+      <div style="display: flex">
+        <div style="overflow: auto; width: 100%">
+          <!-- Just used p-inputtext to keep styling consistent -->
+          <div class="p-inputtext" style="overflow: auto">
```
loneil, line 8: Should stay away from inline styles as much as we can. App uses https://www.primefaces.org/primeflex/ which provides framework classes for these if needed 

https://www.primefaces.org/primeflex/flex
https://www.primefaces.org/primeflex/overflow
https://www.primefaces.org/primeflex/width

(or apply with a named class in the style block of the component or something).

loneil (approved):
Great stuff, will be very helpful for debugging/advanced usage with swagger

PR #671, opened 2023-06-15
Fix public fetch of OCA list endpoint

loneil:
Don't filter if root profile (no tenant token provided). If the endpoint is called as intended publicly, do not filter on a DID.
So this will allow (as designed)
- You to call /oca to get the directory of public OCA-cred associations from Traction as anybody
- You to call it authenticated with your token to get your OCAs that you manage.

Also adjust (and refactor to make it less verbose to do) the Tenant UI to call the OCA endpoint unauthed when appropriate

There's a separate weirdness here being investigated where the base wallet on deployed (but not local docker) Traction instances is coming back with a public DID that was exposing this. Unrelated, but will look into that more. Regardless, the auauthed call to this endpoint shouldn't be looking for a DID to compare anyways.

### DIFF
--- a/plugins/traction_innkeeper/traction_innkeeper/v1_0/oca/oca_service.py
+++ b/plugins/traction_innkeeper/traction_innkeeper/v1_0/oca/oca_service.py
@@ -127,7 +127,7 @@ class OcaService:
         is_root_profile = issuer_profile == self.profile
         public_info = await self.get_public_did_info(issuer_profile)
         tag_filter = self.build_tag_filter(schema_id, cred_def_id)
-        post_filter = {} if is_root_profile == True else self.build_post_filter(public_info)
+        post_filter = {} if is_root_profile else self.build_post_filter(public_info)
         records = []
         if is_root_profile or public_info:
             self.logger.info(f"post_filter = {post_filter}")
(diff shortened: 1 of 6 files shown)


plugins/traction_innkeeper/traction_innkeeper/v1_0/oca/oca_service.py
```diff
@@ -127,7 +127,7 @@ async def list_oca_records(
         is_root_profile = issuer_profile == self.profile
         public_info = await self.get_public_did_info(issuer_profile)
         tag_filter = self.build_tag_filter(schema_id, cred_def_id)
-        post_filter = self.build_post_filter(public_info)
+        post_filter = {} if is_root_profile == True else self.build_post_filter(public_info)
```
usingtechnology, line 130: 👍

plugins/traction_innkeeper/traction_innkeeper/v1_0/oca/oca_service.py
```diff
@@ -127,7 +127,7 @@ async def list_oca_records(
...
         records = []
         if is_root_profile or public_info:
```
usingtechnology, line 132: kind of inconsequential, but could change this logic.

usingtechnology (approved):
I like it; simple way to address it.

plugins/traction_innkeeper/traction_innkeeper/v1_0/oca/oca_service.py
loneil: Oh yeah thanks for catching, left `== True` in there while debugging

Issue #558, opened 2023-04-26
Postgres tuning for high request volume


esune:
As per conversation with @usingtechnology :

It appears that Postgres might not be able to handle swarms of connections by default: the issue was uncovered while attempting to create several tenants programatically. When hitting 20 concurrent requests (each tenant will create its db schema), Postgres throws the following error:
```
sorry, too many clients already
```

The issue was overcome by using a custom `max_conns.sql` with the following contents:
```
ALTER SYSTEM SET max_connections = 500;
```

In docker, the mount path is as outlined in the following snippet (might be different in k8s/OpenShift):
```
wallet-db:
    image: postgres:15
    environment:
      - POSTGRES_USER=DB_USER
      - POSTGRES_PASSWORD=DB_PASSWORD
    ports:
      - 5433:5432
    volumes:
      - wallet-db-data:/var/lib/pgsql/data
      - ./max_conns.sql:/docker-entrypoint-initdb.d/max_conns.sql    
    networks:
      - app-network
```

Acceptance Criteria:
- Add custom `max_conns.sql` configuration file/settings to the Postgres deployment used as dependency by Traction.
- 
-The better way to set the max connections on a Postgres server is to use the Postgres configuration files. We do this with the BC Registries databases. It was done specifically for the wallet (secure storage) in that case. Example here https://github.com/bcgov/von-bc-registries-agent-configurations/blob/main/openshift/templates/db/config/postgresql-cfg/dbconnections.conf. It might be an idea to adopt the configuration as standard for any wallet databases (secure storage).

swcurran:
Should we worry about this with the Mediator @WadeBarnes ?  

Also — are there ways to monitor logs and trigger a notification when a magic string like `sorry, too many clients already` is seen?

WadeBarnes:
The better way to set the max connections on a Postgres server is to use the Postgres configuration files.  We do this with the BC Registries databases.  It was done specifically for the wallet (secure storage) in that case.  Example here https://github.com/bcgov/von-bc-registries-agent-configurations/blob/main/openshift/templates/db/config/postgresql-cfg/dbconnections.conf.  It might be an idea to adopt the configuration as standard for any wallet databases (secure storage).

I'll have to look into how we can be alerted to the issue.

esune:
@i5okie so you are aware of this - the configuration file could be applied to all our postgres deployments in theory

hiteshgh:
waiting for HELM chart to get done and we can then test it

PR #672, opened 2023-06-20
Traction Helm Chart Refresh

i5okie:
# Helm Chart Refresh

- Move environment-specific values files out of the `traction` and `tenant-ui` chart directories to `deploy` directory.
- Update GHA workflows to reflect the changes
- Remove image version/buildtime metadata and resources in helm charts

## Traction
- Refactor helm chart templates, helpers, values
- Add autoscaling enabled via a flag in values
- Add `ingress` enabled via a flag in values. This allows the chart to be installed on a common Kubernetes cluster.
- Disable creating Openshift `route` objects by default for the above reason
- Simplify Acapy configuration by generating configmap with `argfile.yml` and `plugin-config.yml` customizable in Values file
- Improve secrets generation (Generated once, reused on `helm upgrade`)
- Allow Postgres configuration and credentials to be passed via existing secrets or individual values
- Add `max_connections = 400` to postgres configuration (Addresses https://github.com/bcgov/traction/issues/558)
- Annotate values file with metadata
- Generate `README.md` with Bitnami [Readme Generator For Helm](https://github.com/bitnami-labs/readme-generator-for-helm)

## Tenant-UI
- Rename chart
- Refactor helm chart templates, helpers, values
- Use `bitnami/common` helm library chart
- Add `ingress` enabled via a flag in values. This allows the chart to be installed on a common Kubernetes cluster.
- Disable creating Openshift route objects by default for the above reason
- Create an `innkeeper` secret from provided existing secret, or by providing the individual values. (Generated once, reused on `helm upgrade`)
- Annotate values file with metadata
- Generate `README.md` with Bitnami [Readme Generator For Helm](https://github.com/bitnami-labs/readme-generator-for-helm)


charts/traction/templates/acapy/deployment.yaml
```diff
@@ -0,0 +1,178 @@
...
+           "-c",
+           "{{ include "acapy.registerLedger" . }}"
+          ]
+          env:
+            - name: ACAPY_WALLET_SEED
+              valueFrom:
+                secretKeyRef:
+                  name: {{ template "acapy.fullname" . }}
+                  key: seed
+      containers:
```
esune, line 48: @usingtechnology is this container still needed? I assume it is what was being used to register DIDs on the ledger, but we switched to using the endorser request instead - correct?

deploy/tenant-ui/values-int.yaml
```diff
@@ -0,0 +1,29 @@
+fullnameOverride: "int-tenant-ui"
```
esune, line 1: I assume this file will go once the new deployment structure is implemented, correct?

charts/traction/templates/acapy/deployment.yaml
i5okie: I wasn't aware of the switch. Was just moving it out of here https://github.com/bcgov/traction/blob/a3e68e952ad1baf2489377f06f4790054a9a7beb/charts/traction/templates/acapy_deployment.yaml#L39 into its own container.

esune: Np, just leave it there for now - wanted to confirm with @usingtechnology so we can pull this out in the nxt iteration

usingtechnology: i would assume it is still needed (this is a new layout for me)... unless you are starting on multiledger? in which case i guess you wouldn't want to register? that API that is called may not be valid for all ledgers anyway

esune (approved):
👍🏻

PR #676, opened 2023-06-28
feat: reservation OIDC integration

amanji:


loneil:
I think the outlined text here could be better as something like "or Request access with"
as it's not that you're **signing in** with a wallet key or an IDIR at this stage. Probably could cause confusion with that current text

image

amanji:
> I think the outlined text here could be better as something like "or Request access with" as it's not that you're **signing in** with a wallet key or an IDIR at this stage. Probably could cause confusion with that current text
> 
> image

Will update to reflect that wording

loneil:
I'm getting a 404 on checking the status of my reservation, seems to be trying to call to Tenant UI URL with the reservations path

image

services/tenant-ui/frontend/src/store/reservationStore.ts
```diff
@@ -13,7 +13,9 @@ export const useReservationStore = defineStore('reservation', () => {
   // A raw api call without using the interceptors from the acapyApiStore
   // Needed for the open call to reservation at this point
   const api = axios.create({
-    baseURL: config.value.frontend.tenantProxyPath,
+    baseURL: config.value.frontend.showOIDCReservationLogin
+      ? '/'
+      : config.value.frontend.tenantProxyPath,
```
loneil, line 18: Think the goal would be just to proxy through the reservation POST rather than all the ones in the store, see comment about 404)

amanji: I can create a separate method altogether but thought I’d re use what was there.

loneil (approved):
Awesome stuff. Can use my IDIR to reserve and create tenants, can still do the innkeeper stuff with OIDC as before too.

Might as well leave the default to show OIDC for now. Will need some helm values. I will create a ticket for that.

Think this will need some further UX revising, but also along with all the existing reservation flow again as well.

PR #695, opened 2023-07-10
Update workflows to fix deployments

i5okie:
- Update `build tenant-ui` and `build acapy` actions to output only the image tag instead of the full image name with registry, repository, and tag. (Should fix the currently failing dev deployment workflow)
- Rename workflows

### DIFF
--- a/.github/actions/build_acapy/action.yaml
+++ b/.github/actions/build_acapy/action.yaml
@@ -79,6 +79,6 @@ runs:
     - id: values  
       shell: bash
       run: |
-        echo "image_tag=${{ fromJSON(steps.meta.outputs.json).tags[0] }}" >> $GITHUB_OUTPUT
+        echo "image_tag=${{ inputs.ref != '' && inputs.ref ||  env.GITHUB_REF }}${{ inputs.ref }}" >> $GITHUB_OUTPUT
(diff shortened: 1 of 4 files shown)


.github/actions/build_acapy/action.yaml
```diff
@@ -79,6 +79,6 @@ runs:
     - id: values  
       shell: bash
       run: |
-        echo "image_tag=${{ fromJSON(steps.meta.outputs.json).tags[0] }}" >> $GITHUB_OUTPUT
+        echo "image_tag=${{ inputs.ref != '' && inputs.ref ||  env.GITHUB_REF }}${{ inputs.ref }}" >> $GITHUB_OUTPUT
```
WadeBarnes, line 82: I'm not clear on what you're expecting as the image tag.  `inputs.ref` is not required and it looks like you could include it twice.

i5okie: I took that straight from here: https://github.com/bcgov/traction/blob/main/.github/actions/build_acapy/action.yaml#L46
I just want the tag. not the full path to the image with the tag.
Originally the action outputs, for example, `ghcr.io/bcgov/traction-plugins-acapy:sha-18c0499` I just want `sha-18c0499` or whatever the image gets tagged with (the ref that gets passed to the action).

WadeBarnes (approved):

PR #700, opened 2023-07-11
Matomo tracker setup

### LINKED ISSUE #646: Add optional analytics script import
As outlined in https://github.com/bcgov/DITP-DevOps/issues/79, it may be useful to know where users of the system are coming from.

Acceptance Criteria:
- Add optional import of analytics script (e.g.: Matomo, Google Analytics) to be included.
- Evaluate which pages should use the script: it might be required only on the reservations page and tenant UI login, but it might provide insight on other app functionality as well

Gavinok:
Resolves #646 by conditionally importing the Matomo tracker code if a URL for the Matomo server is set.


services/tenant-ui/frontend/src/main.ts
```diff
@@ -37,6 +37,22 @@ async function loadApp() {
       after((result) => {
         console.log('configuration loaded from server.');
         console.log(result);
+        // Matomo Setup
```
loneil, line 40: Does this need to be done in the callback handler? Or could it just be done below after the unsubscribe, using the configStore getter to check the value?

It makes a bit more sense to me to leave this action handler to specifically loading the config, and then if using it to set up matomo doing that below with the other setup steps.

Minor though, fine to handle it here if necessary for now.

services/tenant-ui/frontend/src/matomoSetup.ts
```diff
@@ -0,0 +1,18 @@
...
+	_paq.push(['setTrackerUrl', u + 'matomo.php']);
+	_paq.push(['setSiteId', 1]);
+	var d = document,
+		g = d.createElement('script'),
+		s = d.getElementsByTagName('script')[0];
+	g.type = 'text/javascript';
+	g.async = true;
+	g.src = u + 'matomo.js';
+	//@ts-ignore
+	s.parentNode.insertBefore(g, s);
```
loneil, line 16: does `parentNode?.` fix the ts complaint on this line?

services/tenant-ui/frontend/src/main.ts
```diff
@@ -37,6 +37,22 @@ async function loadApp() {
...
+        const MATOMO_URL: string = result?.frontend.matomoUrl;
+        if (MATOMO_URL) {
+          // @ts-ignore
```
loneil, line 43: Tried locally and I don't see it needing this ts-ignore? Builds this part for me.

services/tenant-ui/frontend/src/matomoSetup.ts
```diff
@@ -0,0 +1,18 @@
+//@ts-ignore
+var _paq = (window._paq = window._paq || []);
```
loneil, line 2: Not pretty but can handle the matomo added `_paq` with
`const _paq = ((window as any)._paq = (window as any)._paq || []);`

Does that break anything?

loneil (changes requested):
Tried a few things locally and can pass build and lint.
Hopefully not breaking any intended behaviour if we do these?

image


image

services/tenant-ui/frontend/src/main.ts
Gavinok: Thanks, this was a hold over from a previous change and I forgot to remove the ts-ignore. eslint was kind enough to point it out

services/tenant-ui/frontend/src/matomoSetup.ts
Gavinok: I just made `s` of type any. This way we will at least get an error if this some how went wrong. We handle the error in the main.ts

services/tenant-ui/frontend/src/matomoSetup.ts
Gavinok: That works great, thanks Lucas

services/tenant-ui/frontend/src/main.ts
Gavinok: Makes a lot of sense. Not much point in complicating this.

services/tenant-ui/README.md
```diff
@@ -90,3 +90,15 @@ The Tenant UI Inkeeper functionality can be configured to log in with either (or
 To set up the OIDC provider of your choice, add configuration values in your deployment to match the `frontend.innkeeperOidc` fields for a auth code grant client,a nd configure the `server.oidc` fields appropriately to veify the JWKS for tokens from that client. 
 
 As well, the Innkeeper secret must be available to the Tenant UI server, this is set in `server.innkeeper` configuration...
+
+## Configuring Matomo
+
+If you would like to use Matomo for tracking you can set the `FRONTEND_MATOMO_URL`
+environment variable as a one time override to the url exposed by your Matomo instance.
+If you would like a more permanent solution set `frontend.matomoUrl` in
+[custom-environment-variables.json](../config/custom-environment-variables.json)
```
loneil, line 99: Wouldn't say it's a one-time solution or permanent solution (in fact, no values should be explicitly set in custom-environment-variables, that file is for pulling environment variables into node config)

Would just leave it as "If you would like to use Matomo for tracking you can set the `FRONTEND_MATOMO_URL` environment variable as exposed in [custom-environment-variables.json](../config/custom-environment-variables.json)"

loneil (dismissed):
Looking good to me!

services/tenant-ui/frontend/src/main.ts
```diff
@@ -51,6 +51,22 @@ async function loadApp() {
...
+  const MATOMO_URL: string = configStore.config.frontend.matomoUrl;
+  if (MATOMO_URL) {
+    import('./matomoSetup')
+      .then((m: { setup: (url: string) => void }) => {
+        console.log(m.setup(MATOMO_URL));
+        console.log('initialized Matomo');
+        console.log(`${MATOMO_URL}`);
+      })
+      .catch((error) => {
+        console.log(error);
```
amanji, line 64: Should log errors with `console.error`, I think, if we're not handling the error and just logging it

services/tenant-ui/frontend/src/main.ts
```diff
@@ -51,6 +51,22 @@ async function loadApp() {
...
+      });
+  } else {
+    console.log('Matomo not configured');
```
amanji, line 67: Can use `console.warn` here

Gavinok: Thanks Akiff, appreciate it!

loneil (approved):

PR #701, opened 2023-07-11
CI/CD Updates and Fixes

i5okie:
- Cleanup and update values files.
- Add missing values to dev/test/prod values files.
- Reduce default resource requests and limits for deployments.


loneil (approved):
Looks good from a config standpoint.
I know the cpu and mem limits were set fairly high in the past because aca-py was using a **lot** of resources.
So not sure if these lower limits have been tried out in the other DITP environment or checked out? We can always give it a try and see.

Maybe @usingtechnology remembers or has thoughts on resourcing needs being higher than this?

PR #702, opened 2023-07-11
Unit Testing [Draft]

jamshale:
This addresses issue 419 of the backlog.

In order to enable the ability to develop using TDD and have code coverage in the frontend 2 major things needed to be done. One thing was to mock the external API in order to test the logic in pinia stores. The other thing was then mocking the pinia stores to allow ui components to mount during testing and allow the tester to change states. The testing files are kept in a /test directory outside of src and mirrors the structure. This is to prevent snapshots and tests polluting the src directory.

To mock the API Mock Service Worker was used ( https://mswjs.io/ ):

- The responses are maintained in the directory /test/__mocks__/api/responses. 
- The responses are divided by store in separate files and name spaced with an index file. 
- The routes follow the same pattern in /test/__mocks__/api/routes. 
- The routes are separated into successful and failed routes. The mock server is then setup in /test/setupApi.ts. The success routes are loaded initially and the server is reset after every test.
- The responses where copied from the browser console or from the API documentation. They could be slimmed down dramatically to only have tested properties but I decided to leave to full responses. 
- Store tests are a bit slow though due to reloading the routes for every test. Less console logging and small response payloads would likely improve the speed.

To allow testing of UI components:

- The tricky part here was mocking the pinia storeToRefs function with the store data while still allowing pinia to work as normal so the createTestingPinia plugin can be used ( https://pinia.vuejs.org/cookbook/testing.html#unit-testing-components )
- A test/setupGlobalMocks.ts file was created where mocks are loaded for every test file.
- Here pinia is copied and the storeToRefs function is mocked with all the store mocks. These are found in /test/__mocks__/store. The files are also namespaced with an index file.
- The stores are all mocked here as well just to avoid doing them in test files.
- Not all the store mocks are completed as they haven't been needed yet. But the files are still initialized and loaded for when they are.

My goal here wasn't to test everything but many tests were created to discover patterns. Most of the tests are deliberately simple and try to not be fragile. For components unlikely to change snapshots were used. 

Even with these simple tests coverage has gotten pretty good. Although the quality and detail of the tests in critical areas could be improved a lot. 

image

One file in src was changed `@/helpers/constants.ts`.


jamshale:
Installs are failing because of 403 error ?

loneil:
Thank for for this appreciated contribution of thorough frontend component and pinia code coverage @jamshale 

One note on efforts of component testing is that there's probably some significant UX review and adjustment of the frontend app due at some point so template structure and component makeup/organization could be altered a lot at some point.

For forms, Primevue at the time was using Vuelidate in their documentation but since has switched to documenting with VeeValidate so I could see changing form validation at some point.
Also form input template setup for validation is far too verbose and there's a need to DRY out how form inputs (extracting to custom components or composables or something). So again, likely some structure change to those elements that are tested here possible in the future.

The 403 error is due to not allowing pull requests from forks to deploy and run any code in the Openshift environment the GHAs in this repo are pointing to or expose secrets.
We could do things like have a approved contributor merge forked PRs into a branch on the main repo after review to deploy a working environment on OCP for testing. Have not needed to flesh out that workflow and we are adjusting CI/CD processes at the moment as well.

For this one since it's specifically tests only in the Tenant UI, as the tests run successfully on this PR, and I have pulled locally and can build/lint/test I'm inclined to just merge this accepted PR.
Will let @amanji or anyone else have any thoughts before that as well.

loneil (approved):

loneil:
(associating to new issue 707 as 419 is an epic to categorize Tenant UI testing, not just specifically Vue frontend)

services/tenant-ui/frontend/src/helpers/constants.ts
```diff
@@ -62,13 +62,13 @@ export const API_PATH = {
...
     `/innkeeper/reservations/${id}/deny`,
 
-  ISSUE_CREDENTIAL_RECORDS: 'issue-credential/records',
-  ISSUE_CREDENTIAL_RECORD: (id: string) => `issue-credential/records/${id}`,
+  ISSUE_CREDENTIAL_RECORDS: '/issue-credential/records',
+  ISSUE_CREDENTIAL_RECORD: (id: string) => `/issue-credential/records/${id}`,
   ISSUE_CREDENTIAL_RECORDS_SEND_OFFER: (id: string) =>
-    `issue-credential/records/${id}/send-offer`,
-  ISSUE_CREDENTIALS_SEND_OFFER: 'issue-credential/send-offer',
+    `/issue-credential/records/${id}/send-offer`,
```
amanji, line 68: Will changing these impact the API calls? OR have they been tested as well?

services/tenant-ui/frontend/test/__mocks__/api/routes/config.ts
amanji, line 1: While these are awesome, I wonder if this testing actually belongs in ACA-Py or the traction plugins themselves. The only API boundary we really need to test is the Node.js backend.

Same comment for all of the tests against `API_PATH` endpoints.

services/tenant-ui/frontend/src/helpers/constants.ts
jamshale: I did test this by putting an invalid path and then with/without the slash. I did that a long time ago though so I can test it again.

services/tenant-ui/frontend/test/__mocks__/api/routes/config.ts
jamshale: These would be useful if the tenant-ui wanted to do more advanced features or error handling with the responses. 
Right now the responses aren't really being used. I think there was a couple tests where the response data changed the code path where I was using it and then decided to have the actual responses just for completeness.
They aren't testing the responses. They could be simplified dramatically if we wanted.

jamshale: Oh, and a few tests just are used as assertions that the response is what is expected. These aren't very useful as it's just checking the mocked response makes it into the store. Might be handy if more logic gets added to the acapyApi store one day.

amanji:
I don’t have any reservations. I think this has been some very comprehensive work that will serve as a great example to other projects.

amanji (approved):

PR #715, opened 2023-07-18
Do not disable name and email for OIDC reservation

loneil:
Take away the disabling of the name and email fields for now in the reserve form when using OIDC access.
Had consumers with IDIRs in the platform Keycloak that did not have email addresses unable to continue with reservation.


esune (approved):
I am still wondering about disabling fields only if they are not set with an OIDC-provided value...

Issue #698, opened 2023-07-10
[Traction Helm Chart] Acapy deployment fails with empty secret values


i5okie:
Changing the `generated` boolean and running `helm upgrade` overrides the existing secret with an empty string.
This results in failed deployment, and loss of previously generated secret keys, if `adminApiKey.value`, `pluginInnKeeper.walletkey`, and `pluginInnKeeper.tenantid` are empty.

## How to reproduce
Install Traction with the following values overrides:
```
...
acapy: 
  secret:
    adminApiKey:
      generated: true
      value: ""
    pluginInnkeeper:
      generated: true
      walletkey: ""
      tenantid: ""
...
```

Upgrade Traction with the updated values overrides:
```
...
acapy: 
  secret:
    adminApiKey:
      generated: false
      value: ""
    pluginInnkeeper:
      generated: false
      walletkey: ""
      tenantid: ""
...
```

The secrets `traction-api`, and `traction-acapy-plugin-innkeeper` with generated values, will be overridden with empty string values.

PR #725, opened 2023-07-20
[Helm chart] Improve secret creation on upgrade

i5okie:
This PR addresses https://github.com/bcgov/traction/issues/698

- Updated API secret
- Updated innkeeper secret

New behavior:
- Helm install/upgrade fails if `.Values.acapy.secret.pluginInnkeeper.generated` is set to `false` and  `.Values.acapy.secret.pluginInnkeeper.tenantid` and/or `.Values.acapy.secret.pluginInnkeeper.walletkey` are emty.
- Helm install/upgrade fails if `.Values.acapy.secret.adminApiKey.generated` is set to `false` and  `.Values.acapy.secret.adminApiKey.value` is empty
- Running helm upgrade does not overwrite existing API secret value when `.Values.acapy.secret.adminApiKey.generated` is set to `true` and API secret exists with populated `adminApiKey` key.
- Running helm upgrade does not overwrite existing pluginInnKeeper secret values when `.Values.acapy.secret.pluginInnkeeper.generated` is set to `true` and pluginInnKeeper secret exists with populated `tenantid` or  `walletkey` keys.

Known issues:
- If a secret exists and the value is empty, the helm upgrade command still succeeds, but the empty key is removed.
     - Re-running helm upgrade with `.generated` set to `true` succeeds but the key is not re-created.
     - Re-running helm upgrade with `.generated` set to `false` succeeds but the key is re-created with the value read from Values.


loneil:
Think it looks good if I'm following along. Should be able to tell if something breaks after deployment to existing envs.

esune (approved):

PR #729, opened 2023-07-21
Try out 0.9.0rc

loneil:
1. Update acapy image to 0.9.0rc `https://github.com/hyperledger/aries-cloudagent-python/pkgs/container/aries-cloudagent-python/111618235?tag=py3.9-0.9.0-rc0`
2. Update plugin dependencies to `https://pypi.org/project/aries-cloudagent/0.9.0rc0/`

### DIFF
--- a/plugins/basicmessage_storage/poetry.lock
+++ b/plugins/basicmessage_storage/poetry.lock
@@ -2,99 +2,99 @@
 
 [[package]]
 name = "aiohttp"
-version = "3.8.4"
+version = "3.8.5"
 description = "Async http client/server framework (asyncio)"
 category = "main"
 optional = false
 python-versions = ">=3.6"
 files = [
-    {file = "aiohttp-3.8.4-cp310-cp310-macosx_10_9_universal2.whl", hash = "sha256:5ce45967538fb747370308d3145aa68a074bdecb4f3a300869590f725ced69c1"},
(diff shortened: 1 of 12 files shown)


loneil:
Tenant
9ba4c87b-e991-4d4f-827f-989006415b71
9ba4c87b-e991-4d4f-827f-989006415b71

loneil:
Current state here, back to creds ending up abandoned 

Different error message than `0.8.2` askar update attempt though (which was working with `0.8.3pre` image)
When issuing/accepting, get `"error_msg": "Error creating credential. Missing tails file path.",`

But the tails files are getting created there this time
image

Need more investigation

esune:
It turns out the `multitenant_provider` was pulling-in an old version of `indy-credx` and that was causing a dependency mismatch between ACA-Py and the plugin. Updating the plugin dependency to match ACA-Py fixes issues and I have been able to (on localhost so far):
- Connect to endorser
- Publish my DID
- Connect to a mobile wallet
- Create a schema
- Create a CredDef with revocation enabled
- Issue a credential
- Revoke a credential

I think this is probably good enough as smoke test to go ahead and approve/wrap-this up?

usingtechnology (approved):
LGTM - the only explicit indy-credx is in multitenant_provider, and all the other plugins should be ok with the new lock files that pick up the correct version due to the ACA-Py version.

PR #743, opened 2023-07-31
Feature flags 717

popkinj:
New toggle attribute within the `local.json` and [`default.json`](https://github.com/bcgov/traction/blob/feature-flags-717/services/tenant-ui/config/default.json#L10) files to turn all writable features (aside from contacts) on/off. 

Here is the full list of features effected:
- Remove function -> Copy Schema
- Remove function -> Create Schema
- Remove function -> Create a Message from messages page
- Remove function -> Create a Message from message sidebar in the connections page
- Remove function -> Edit and Delete Contact
- Remove function -> Add/Update OCA Associations
- Remove function -> All buttons in the CredentialsTable.vue
- Remove function -> Verifications; Create Presentation Request, Delete and Copy
- Remove function -> Create Credential Definition
- Remove function -> Create Presentation Request... and delete and copy in the table


popkinj:
I'm seeing some build errors.... Assuming it has something to do with the Helm chart still needing an update. Will look into that now.

loneil (commented):
`default.json` is not sufficient to control deployment values. That is just the hard-coded built setting if the environmental setting is not there.
Need to add same key to `custom-environment-variables.json` (uses standard [Node-Config hierarchy](https://github.com/node-config/node-config/wiki/Configuration-Files)) and add an environment variable name that can be set by the Helm Charts.

For example, see how in 
- `default.json` there is the default value for `showInnkeeperAdminLogin` 
- And in  `custom-environment-variables.json` that maps it to an env var `showInnkeeperAdminLogin": "FRONTEND_INNKEEPER_SHOW_ADMIN",`
- Then to get that variable set from helm, in `config.yaml` it sets up a helm chart mapping
`FRONTEND_INNKEEPER_SHOW_ADMIN: {{ .Values.oidc.showInnkeeperAdminLogin | quote }}`
- which is then specified per-environment in the corresponding values files, such as `values-pr.yaml`, `values-production.yaml` etc.

Same thing should be done here with the new `showWritableComponents`



Build errors aren't related to Helm values, it's Unit Tests failing. Check reason for failures with "details" link in PR
image
(or run `npm run test` locally in the frontend folder)
Someone else from Quartech PR'd a lot of unit test coverage so I'd assume changes like taking away buttons could trigger failures (though default value being true here I'm not sure how would affect UTs, might need mocks or something).
I haven't had a chance to dig much into the unit test snapshot stuff with ViteTest. Akiff knows more and may be able to help, or skip specific tests if they are getting in the way too much for now.

services/tenant-ui/frontend/src/components/connections/Connections.vue
```diff
@@ -45,13 +45,17 @@
...
             icon="pi pi-trash"
             class="p-button-rounded p-button-icon-only p-button-text"
             :disabled="deleteDisabled(data.alias)"
             @click="deleteContact($event, data.connection_id)"
           />
-          <EditContact :connection-id="data.connection_id" />
+          <EditContact
+            v-if="config.frontend.showWritableComponents"
+            :connection-id="data.connection_id"
+          />
```
loneil, line 58: Not sure we want to remove connection management functionality? Not specified in ticket so maybe a quick question for Emiliano.

services/tenant-ui/frontend/src/components/oca/Oca.vue
```diff
@@ -20,7 +20,7 @@
       <template #header>
         <div class="flex justify-content-between">
           <div class="flex justify-content-start">
-            <CreateOca />
+            <CreateOca v-if="config.frontend.showWritableComponents" />
```
loneil, line 23: I would consider leaving OCA management for production "read-only" cases. It's not related to Line of Business data or credential management. It's almost more of a Tenant Setting thing.
This allows users to set a specified OCA bundle with a credential definition they manage. So a "traction ecosystem" can be built up where known holders can determine which OCA from a known issuer they are meant to display. Something that seems like settings management through the Tenant UI to me I guess.

Could be more discussion, simple to change in or out one way or another with this new flag.

services/tenant-ui/frontend/src/components/connections/Connections.vue
esune: I think connection management should stay, as connections are often handled out-of-band (i.e.: by hand) before two agents can communicate.

services/tenant-ui/frontend/src/components/oca/Oca.vue
esune: I am thinking we hide this until we have a better "common" OCA set-up: this wouldn't affect outbound credentials, but only credentials stored in the wallet and while the UI would make them prettier I am not too sure users would go in and associate a bundle to their credentials manually. Plus it can be easily reactivated if/when deemed necessary, as you mentioned.

jamshale:
The unit tests are failing because I wrote some tests that was preventing features that I thought were important from being removed. They are basically doing what they are supposed to.

To fix them:

- I removed the delete connections tests in `/components/connections/Connections.test.ts`
- The `OfferCredential` component check in issuedCredentials component because it's hidden now.
- The `CreateMessage` component check in Messages component.
- The `CreateRequest` and `DeleteExchangeRecord` components check in Presentations.

I could fix it for you but haven't got direct push access to the repo yet. I still need to do a fork.

I know it's a bit of a change but having the tests run when you are making changes is a good way to do continuous testing. These tests would have failed when you saved the file.
Maybe the tests were a bit to intrusive. But i think they were valuable if someone hid these components by mistake.

popkinj:
> The unit tests are failing because I wrote some tests that was preventing features that I thought were important from being removed. They are basically doing what they are supposed to.
> 
> To fix them:
> 
> * I removed the delete connections tests in `/components/connections/Connections.test.ts`
> * The `OfferCredential` component check in issuedCredentials component because it's hidden now.
> * The `CreateMessage` component check in Messages component.
> * The `CreateRequest` and `DeleteExchangeRecord` components check in Presentations.
> 
> I could fix it for you but haven't got direct push access to the repo yet. I still need to do a fork.
> 
> I know it's a bit of a change but having the tests run when you are making changes is a good way to do continuous testing. These tests would have failed when you saved the file. Maybe the tests were a bit to intrusive. But i think they were valuable if someone hid these components by mistake.

All good @jamshale ... At least we know they're doing what they're supposed to.
I'll have a look at doing the test updates in my branch. When done I'll let you know so you can have a look. :)

esune:
@popkinj @jamshale if the end to end test can be aware of the flag and check things accordingly it would be the best of both worlds.

jamshale:
> @popkinj @jamshale if the end to end test can be aware of the flag and check things accordingly it would be the best of both worlds.

@esune You're right. The better way than removing the tests is to set the showWritableComponents flag in configStore mock in `/test/__mocks__/store/config.ts` to true. There are other flags there. Because this is undefined is why the UI components aren't showing.

You could toggle it to false during a test and make sure they are hidden to test this feature.

services/tenant-ui/frontend/src/components/connections/Connections.vue
popkinj: Done.

services/tenant-ui/frontend/src/components/oca/Oca.vue
popkinj: OK... the button is hidden now for everyone.

services/tenant-ui/frontend/src/components/oca/Oca.vue
```diff
@@ -20,7 +20,7 @@
       <template #header>
         <div class="flex justify-content-between">
           <div class="flex justify-content-start">
-            <CreateOca />
+            <!-- <CreateOca v-if="config.frontend.showWritableComponents" /> -->
```
loneil, line 23: Is this supposed to be commented out, rather than just `v-if`'d like all the others?

esune: Sorry, my comment was misleading. I would hide it in the "production" view, it can stay as a sandbox/development feature though.

popkinj: My bad.....
K. I'll fix that.

jamshale:
I don't have a problem with removing the tests but i think you should declare the flag as false in `/test/__mocks__/store/config.ts` the config store mock instead of leaving it undefined.

loneil:
Yeah I'd rather see tests for the on/off cases for this flag (and declare it in the test data) but if removing them can get this ticket done then can move forward and address later.

deploy/tenant-ui/values-production.yaml
```diff
@@ -9,6 +9,7 @@ traction:
 oidc:
   active: true
   showInnkeeperAdminLogin: true
+  showWritableComponents: true
```
loneil, line 12: THink the goal is to be `false` for production

popkinj:
> I don't have a problem with removing the tests but i think you should declare the flag as false in `/test/__mocks__/store/config.ts` the config store mock instead of leaving it undefined.

Done. :+1:

popkinj:
> Yeah I'd rather see tests for the on/off cases for this flag (and agree, declare it in the mock test data) but if removing them can get this ticket done then can move forward and address later.

My thoughts exactly. I'll make another ticket for this.

popkinj:
> 

Will bring these tests back in another PR. :)

loneil:
Testing value false helm flag

loneil (changes requested):
Doesn't appear to be working. I set the helm val to false and I can still see the buttons.
Possible string/boolean JS truthiness issue?

image


image

popkinj:
> 58d73890-db96-4bec-992c-b03df2200b16

How did you manage to get quotes around the boolean of false? I'm not seeing them in the config. Hmmm... 🤔

loneil:
> How did you manage to get quotes around the boolean of false? I'm not seeing them in the config. Hmmm... 🤔

The Developer page just prints out the config.

popkinj:
>

popkinj:
Oops didn't mean to close this

services/tenant-ui/frontend/src/components/messages/Messages.vue
```diff
@@ -21,7 +21,10 @@
         <div class="flex justify-content-between">
           <div class="flex justify-content-start">
             <CreateMessage
-              v-if="config.frontend.showWritableComponents"
+              v-if="
+                config.frontend.showWritableComponents === true ||
+                config.frontend.showWritableComponents === 'true'
+              "
```
loneil, line 27: I think this is probably necessary if the Helm charts are injecting a string from the bool in the yaml.
Ideally the helm chart to Node-Config would result in a boolean but I don't know that off the top of my head...

Maybe ideally some util function for this like
```
function isConfigTrue(value: any): boolean {
  return value === true || value === "true";
}
```

so the logic is encapsulated?

see `services\tenant-ui\frontend\src\helpers\index.ts`

popkinj: That looks pretty slick.

popkinj: I just did a quick search and replace. May use the helper in the future. Thanks for the suggestion @loneil

popkinj:
Looks like this is working now. 🎉

loneil (changes requested):
Looks good with the `false` flag working now. Couple minor points:

Think there's a spot or 2 missing the hiding 
- IssuedCredentials.vue, the `DeleteCredentialExchangeButton` and `RevokeCredentialButton` (referenced in issue, revoke is at least)
- Schema Storage `Delete Schema` button.

Verifications page seems to pull the search over the left when the button is gone
image
**Issuance** page behaves correct here so can just copy what that is doing

loneil (approved):
Looks good to me 👍 
Tried flipping the config val manually in OCP console and looks good in both modes now.
Can just flip that flag in `values-pr` back to true now that testing is done and then can merge.

PR #737, opened 2023-07-27
Tenant Issuer Config Related Fixes

shaangill025:
- resolve #607 
- resolve #608 
- resolve #667

PR #748, opened 2023-08-08
Tenant Issuer Config Fixes

### LINKED ISSUE #607: Configuration to create Tenants with default permission values
As a Traction operator, I want a flag to control the issuer approval requirements, so that I can run an instance where Tenants can make themselves issuers without Innkeeper intervention
--

AC:
- There are flags for the Traction instance appropriate plugin that can be set in Helm and Docker
- These flags contain the "default permission values" for the endorser and public did permission settings
- When Tenant creation occurs, the Tenant **Config** sets the configured default for the connect to endorser and create public DID config values. 
IE these ones:
```
  "connect_to_endorsers": {
    "endorser_alias": " ... ",
    "ledger_id": " ... "
  },
  "create_public_did": [
    "string"
  ],
  ```
  can have default values in the plugin configuration.
- If the default values are not present or blank arrays, the tenant does not get those permissions set (and the Innkeeper must manually approve the permissions.
- The Tenant UI still behaves accordingly in the Issuance section

Relates to work previously done in
- https://github.com/bcgov/traction/issues/393
- https://github.com/bcgov/traction/issues/394
- https://github.com/bcgov/traction/issues/570

### LINKED ISSUE #608: Block endorser connection and public DID tenant endpoints if Innkeeper approval is not set
As a Traction operator I want the Innkeeper approval for issuance-setting to be enforced on the plugin endpoints, so that a Tenant can't call them directly to become an issuer.
--

This is handled in the Tenant UI now but needs to be blocked in the API endpoints as well

AC:
- These 2(?) endpoints check whether the Tenant calling them is allowed to call them based on the Innkeeper allowance config.
- https://traction-tenant-proxy-dev.apps.silver.devops.gov.bc.ca/api/doc#/traction-tenant/post_tenant_endorser_connection
- https://traction-tenant-proxy-dev.apps.silver.devops.gov.bc.ca/api/doc#/wallet/post_wallet_did_create
(I think, check these are the endpoints we want to affect when doing this ticket)
- The Tenant UI still works as before (likely no changes needed for the UI in this ticket)
- The API returns an appropriate response code and message telling the caller they can't do the operation

Related to previous work done in 
- https://github.com/bcgov/traction/issues/393
- https://github.com/bcgov/traction/issues/394

### LINKED ISSUE #667: Can't clear the connect_to_endorser and create_public_did fields in config
In the 
PUT /innkeeper/tenants/{tenant-id}/config
https://traction-tenant-proxy-dev.apps.silver.devops.gov.bc.ca/api/doc#/traction-innkeeper/put_innkeeper_tenants__tenant_id__config
endpoint, you can update or set the fields in the arrays, but it doesn't seem to propagate that through if you set it to blank arrays.

See screenshot below request:

image

Then getting the tenant config after that shows the fields have not been updated:

image

shaangill025:
Reopening #737 here as there were issues in the PR development environment after changing `self_issuer_permission` to `auto_issuer` in `TenantRecord`.
 
- resolve #607 
- resolve #608 
- resolve #667


esune (commented):
I was able to create a tenant successfully. The profile page does not allow the tenant to connect to an endorser or publish a did since `auto_issuer` is set to false - that is good. I changed the value of `auto_issuer` to true in the argfile and restarted the container: on startup, aca-py still loads the auto issuer flag as `False`, even if the mounted config file is set to true.

shaangill025:
> I was able to create a tenant successfully. The profile page does not allow the tenant to connect to an endorser or publish a did since `auto_issuer` is set to false - that is good. I changed the value of `auto_issuer` to true in the argfile and restarted the container: on startup, aca-py still loads the auto issuer flag as `False`, even if the mounted config file is set to true.

Is this for `traction_innkeeper`, that is by design? New tenants created after setting `auto_issuer` to true should reflect that change in logs and tenant-ui [tested now locally using `scripts`].

esune:
@shaangill025 you are right. The log statement relates to the innkeeper tenant, which only has admin role and will not connect to an endorser or publish a DID. I also apparently had restarted the wrong pod, after restarting the correct one and creating a new tenant everything works as expected.

esune (approved):
Tested new tenant creation as well as clearing existing settings, everything works as expected 👍🏻

loneil:
👍 Did a quick test of this as well after getting back today by creating a new temp PR with the helm val set to true and all the Tenant UI/Innkeeper behaviour works as I'd expect. Created a new tenant in DEV as well after this was deployed (where the flag is false) and that is still working as expected too. Great stuff!

Probably need to decide if we'd want this flag set to true in pull request or "lower" envs by default in the helm vals as well.

PR #751, opened 2023-08-09
Enter create another attribute similar to + rather than submit schema

### LINKED ISSUE #747: bug: Hitting enter when adding schema attributes creates the schema
Would prefer that enter goes to add another attribute vs. considering the creation of the schema complete.

Nice to have the "+" flow down with the attribute being added.  That's my opinion.

Gavinok:
This PR resolves #747

Now when enter is hit a new attribute is added similar to how `+` works now


swcurran:
I just tried it and it didn’t seem to work.  When I hit “Enter” the schema is created. I want it to add another attribute.

I tried using the deployment URLs for this.  Perhaps we need a zoom session for this?

Gavinok:
@swcurran Just fixed this. Not sure why this wasn't occurring on my local deployment. Now it should work as expected 😄

loneil (approved):
Works for me after latest change, chrome and firefox.

Not sure if the _ideal_ UX would be for it to tab down to the newly created box on enter as well... but the crux of it is to not submit the form, so if that can be done with this one-liner then IMO: 👍 

Actually I kind of like it not tabbing down anyways since if I know I need to add 10 fields I hit enter 10 times at the start and then tab through filling them out personally.

Issue #648, opened 2023-06-06
SPIKE - Expose configuration to Tenant UI


esune:
Acceptance Criteria:
Settings such as ACA-Py version, ledger(s), installed plugins, tails server should be exposed by the Traction plugin so that the Tenant UI (or other client) could consume them.

esune:
@loneil this would be another good candidate to tackle soon, as it would simplify deployments and make front/back ends consistent in behaviour.

loneil:
Yeah I was just thinking I'd like to know exact ACA-Py release in the About page (rather than the config-defined hard-coded value in the JSON).
I was thinking the best way is actually to take the image being pulled into the helm chart/docker env and make an environment variable that config picks up on. So then the exact image tag is known to the Tenant UI.

But then there's all the other config values that would be good to be known from ACA-Py. So maybe both of those things here.

Could be tackled soon yes.

esune:
I think with this spike we want to get to a list of:
- what does the tenant UI need to know exactly
- what does ACA-Py expose already
- what is missing, but is required

We can then come up with a strategy to consolidate.

esune:
Making this high priority: as we move to support production usage of Traction, having less moving parts is going to be essential to maintain operational reliability.

loneil:
@esune I should be good to start on this tomorrow if we want.

PR #780, opened 2023-08-24
Misc Tenant UI refactors

loneil:
Some miscellaneous fixes for the Tenant UI from earlier to get in while between tasks.
Probably don't need to review most files here, most are renames of things.

- Finally complete all the "Contact" > "Connection" renaming so no discrepancies
- Show an indication you are logged in in OIDC reservation (and 'logout')
- Package updates
- Use correct i18n code for JA
- Minor text and style


services/tenant-ui/frontend/src/components/about/Traction.vue
```diff
@@ -3,10 +3,10 @@
     <div class="col-12 md:col-6">
       <strong>{{ $t('about.traction.info') }}</strong>
       <p class="my-0">
-        {{ $t('about.traction.tractionVersion', { version: '0.2.2' }) }}
+        {{ $t('about.traction.tractionVersion', { version: '0.1.1' }) }}
       </p>
       <p class="mt-0">
-        {{ $t('about.traction.uiVersion', { version: '0.2.2' }) }}
+        {{ $t('about.traction.uiVersion', { version: '0.1.1' }) }}
```
loneil, line 9: This should not be hardcoded, but keeping for now. 
Will look at how to address with https://github.com/bcgov/traction/issues/648 among other things

esune: Wonder if this could be a compile-time substitution, potentially - IF we can pull it in from the `package.json`. Otherwise a dynamically-loaded config might be the way.

services/tenant-ui/frontend/src/components/common/LocaleSwitcher.vue
```diff
@@ -9,7 +9,7 @@ import { I18nLocale } from '@/helpers';
 
 const { locale } = useI18n({ useScope: 'global' });
 
-const localeList: Array<I18nLocale> = ['en', 'fr', 'jp'];
+const localeList: Array<I18nLocale> = ['en', 'fr', 'ja'];
```
esune, line 12: Just a note: this might be something we want to look at having dynamically populated based on the locales actually available - if possible.

services/tenant-ui/frontend/src/components/about/Traction.vue
loneil: Yeah the tenant UI one I'd like to pull from package.json.
Traction version I'm not actually sure where we define... could be the release version pulled in to dynamic config

esune (approved):

PR #781, opened 2023-08-25
Add unit tests for showing/hiding ui elements with config flag

jamshale:
Added unit tests for important ui components that get hidden by the showWritableComponents flag.

Can be used as an example for changing the config store values.


loneil (approved):
Looks good 👍

PR #783, opened 2023-08-28
Add browser storage and router middleware for logout/refresh

jamshale:
Features:

- Prevents user needing to login after refresh by saving token and tenant data in local storage.
- Creates router middleware to better handle logout path.
- Creates inactivity timer that will log the user out if they haven't used the app for a configured amount of time (5 mins). This was attached to the Header component.
- Creates a modal countdown to warn the user they are about to be logged out (10 seconds)
- If the users access token expires they receive a toast notification with the reason (This is unlikely with inactivity session)
- If the user closes the app and opens it again they will need to login again. This doesn't prevent a bad agent from stealing the access token but does prevent them just navigating back to the app and not needing to login.
- Added some unit tests. More advanced ones could be added for the inactivity session but don't have the time right now.

Unknowns:

- Didn't test with the OIDC login. I don't anticipate problems but i'm not sure how to set that up right now and don't really have the time to figure it out. Should definitely be looked at before going to prod.

Think I looked at the other comments and didn't see any issues.

Screen-shots:

- inactivity with countdown stopped and then let expire (set to 5 seconds)

[Screencast from 2023-08-30 11:54:31 AM.webm](https://github.com/bcgov/traction/assets/31809382/289cebb4-7c72-4ad5-b823-de9687d3dc36)

- refreshing and exiting the application

[Screencast from 2023-08-30 11:57:06 AM.webm](https://github.com/bcgov/traction/assets/31809382/9ac36f09-c3b3-4781-a5ff-ea09b55ef897)

Not sure why you can't see my mouse on these. Makes it a bit hard to understand what i'm doing.

### DIFF
--- a/services/tenant-ui/config/default.json
+++ b/services/tenant-ui/config/default.json
@@ -8,6 +8,10 @@
     "showInnkeeperAdminLogin": true,
     "showOIDCReservationLogin": false,
     "showWritableComponents": true,
+    "session": {
+      "timeoutSeconds": "600",
+      "countdownSeconds": "30"
+    },
     "oidc": {
       "active": false,
       "authority": "https://dev.loginproxy.gov.bc.ca/auth/realms/digitaltrust-nrm",
--- a/services/tenant-ui/frontend/src/store/tenantStore.ts
+++ b/services/tenant-ui/frontend/src/store/tenantStore.ts
@@ -42,12 +42,19 @@ export const useTenantStore = defineStore('tenant', () => {
   function clearTenant() {
     console.log('> clearTenant');
     tenant.value = null;
+    localStorage.removeItem('tenant');
+    localStorage.removeItem('tenantConfig');
+    localStorage.removeItem('tenantTaa');
+    localStorage.removeItem('tenantEndorserInfo');
+    localStorage.removeItem('tenantEndorserConnection');
     console.log('< clearTenant');
   }
 
   async function getSelf() {
     tenant.value = await fetchItem(API_PATH.TENANT_SELF, '', error, loading);
+    localStorage.setItem('tenant', JSON.stringify(tenant.value));
   }
+
   async function getTenantConfig() {
     tenantConfig.value = await fetchItem(
       API_PATH.TENANT_CONFIG,
@@ -55,6 +62,7 @@ export const useTenantStore = defineStore('tenant', () => {
       error,
       loading
     );
+    localStorage.setItem('tenantConfig', JSON.stringify(tenantConfig.value));
   }
 
   async function getIssuanceStatus() {
@@ -91,6 +99,10 @@ export const useTenantStore = defineStore('tenant', () => {
       .getHttp(API_PATH.TENANT_ENDORSER_CONNECTION)
       .then((res: any) => {
         endorserConnection.value = res.data;
+        localStorage.setItem(
+          'tenantEndorserConnection',
+          JSON.stringify(endorserConnection.value)
+        );
       })
       .catch((err) => {
         endorserConnection.value = null;
@@ -124,6 +136,10 @@ export const useTenantStore = defineStore('tenant', () => {
       .getHttp(API_PATH.TENANT_ENDORSER_INFO)
       .then((res: any) => {
         endorserInfo.value = res.data;
+        localStorage.setItem(
+          'tenantEndorderInfo',
+          JSON.stringify(endorserInfo.value)
+        );
       })
       .catch((err) => {
         error.value = err;
@@ -174,6 +190,7 @@ export const useTenantStore = defineStore('tenant', () => {
       error,
       !loadingIssuance.value ? loadingIssuance : ref(false)
     );
+    localStorage.setItem('tenantPublicDid', JSON.stringify(publicDid.value));
   }
 
   async function registerPublicDid() {
@@ -313,6 +330,7 @@ export const useTenantStore = defineStore('tenant', () => {
       error,
       !loadingIssuance.value ? loadingIssuance : ref(false)
     );
+    localStorage.setItem('tenantTaa', JSON.stringify(taa.value));
   }
 
   async function acceptTaa(payload: object) {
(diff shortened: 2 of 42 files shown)


loneil:
The Tenant token should have an expiry:

image

It's set fairly long in our environments at this time (24 hrs) but should be configurable.

If we're observing that token not adhering to it's expiry (I don't know if in my time I've tried that out) we should raise it with ACA-Py community.

jamshale:
@loneil @esune  
Ok I understand the token a bit better. 24 hours seems ok. I think it should try and delete the local storage if the user closes the tab or window. `The user should need to login again if they close the browser? ie: This shouldn't let the user back in without a login for a day?`

A session activity logout would probably be a good idea along with the checkbox to prevent using local storage at all. Not sure if all of that should be done in one PR or not.

services/tenant-ui/frontend/src/store/tenantStore.ts
```diff
@@ -335,6 +351,42 @@ export const useTenantStore = defineStore('tenant', () => {
...
+      const savedPublicDid = localStorage.getItem('tenantPublicDid');
+      publicDid.value = savedPublicDid ? JSON.parse(savedPublicDid) : null;
+    } catch (err) {
+      console.info('Error processing tenant info from local storage');
+      console.error(err);
+    }
+
+    console.log('< setTenantFromLocalStorage');
+  }
+
```
loneil, line 389: As probably noticed there's a series of calls to ACA-Py that are needed when logging in to get required tenant "profile" information which these come from.

I'm not sure we'd want to save all these to local storage though? My initial thought were it would be best to just store the **token** in local storage. If that's there then on refresh then we don't need to go through the login form, but would fetch these things on app mount/view load/wherever. So more stateless other than the token.

That probably would required some refactoring to have those calls be made outside of the LoginForm.vue. I hadn't thought through entirely what's the best pattern for this, but I think storing them could cause some issues like if something is changed in your config on the backend and you come back to the app later you'd be picking up older configuration information.

loneil: Yeah it's nice to not have the (multiple!) calls for profile things when you refresh. 
Only thing I guess is they'd be local storage sitting in the browser and readable to someone later on whereas the token eventually expires so it's limited attack vector.
But as you said a session timer would be good, and yes I don't think anything being stored isn't public enough anyway.

So yeah we could try it out this way I think.
Note the `isIssuer` flag is used a number of places around the app and it's dependent on some of this profile info so we'd have to check out that works in this paradigm with refreshes even if you're not on a page that explicitly loads any of these.

services/tenant-ui/config/default.json
```diff
@@ -8,6 +8,11 @@
     "showInnkeeperAdminLogin": true,
     "showOIDCReservationLogin": false,
     "showWritableComponents": true,
+    "tokenExpirySeconds": "86400",
+    "session": {
+      "timeoutSeconds": "600",
+      "countdownSeconds": "10"
+    },
```
loneil, line 15: These can (probably should) be added to the environment configuration and then propagated through helm charts. Unless we want to just leave them as hard coded defaults... which for this feature wouldn't be that bad as it's not really something that would probably change too much per environment. (though the expiry might vary actually)

That can definitely be done in a subsequent PR after we try this out, unless you're really keen to take a crack at it (I can show you where things are defined etc), but if so definitely want to rebase in all of Ivan's latest PR as all the helm chart files are recently changed locations.

services/tenant-ui/config/default.json
```diff
@@ -8,6 +8,11 @@
     "showInnkeeperAdminLogin": true,
     "showOIDCReservationLogin": false,
     "showWritableComponents": true,
+    "tokenExpirySeconds": "86400",
```
loneil, line 11: Is this meant to sync up with the ACA-Py token lifespan? If so we can have those pull from the same environmental config I _think_. But that will be totally in the helm chart realm, this default field would remain.

loneil (commented):
Just a couple questions before getting started testing, and we should update the branch to account all the changes from Ivan's environment config stuff (nothing that would cause merge conflicts) just to be safe before testing too much.

loneil:
@jamshale the OIDC login is working (this is on the Innkeeper side), though if I refresh it doesn't have the cached token.

I forget but that OIDC callback logs you in through a slightly different path.
Can search for `_userManager.events.addUserLoaded` and maybe do the same storage calls there. Or leave the OIDC part for later, it's really only relevant here when logging in through IDIR to the Innkeeper portal.

services/tenant-ui/config/default.json
jamshale: Yes it was supposed to link up. I'm not sure if this feature is needed anymore. I added it before I added forcing the user to login again on a new tab/window. If the token expired they would get "randomly" logged out and wanted to tell them why. But now the user would need to use the app for a day straight without the inactivity log out.

I think I may remove this feature.

jamshale: This is removed now

services/tenant-ui/config/default.json
jamshale: I added these variable at 10 mins activity session and 30 seconds countdown for all environments.

jamshale:
> @jamshale the OIDC login is working (this is on the Innkeeper side), though if I refresh it doesn't have the cached token.

I added what I think is the refresh token to the local storage in the oidc login path. I think it should work but I don't have IDIR credentials to try it.

loneil:
> > @jamshale the OIDC login is working (this is on the Innkeeper side), though if I refresh it doesn't have the cached token.
> 
> I added what I think is the refresh token to the local storage in the oidc login path. I think it should work but I don't have IDIR credentials to try it.

FYI, it's not the Oauth refresh token, but just the same type of tenant token as the other login paths use.
The way the Innkeeper OIDC works (if interested in the context 😄 )
- Auth with OIDC provider
- On success callback OIDC access token is sent to Tenant UI **backend** app
- Backend app validates token and **it** has secured access to Innkeeper creds.
- Fetches **ACA-Py** token on behalf of Innkeeper, returns that ACA-Py tenant token to the browser, same as if you would log in with the credentials
- Just use that Tenant token for authentication and calls.
- So the OIDC login and token are only kind of used as a proxy for that very first login step, no refresh token needed, nor is the ODIC access token used again (or stored).

Seems to be working when logging in with my IDIR and refreshing 👍 

I can test out some more and review fuller a bit later.

loneil:
Noticed some multi-tab inconveniences. If I do something like the following
- Open 2 tabs
- Log in one with a tenant
- Log in another with the Innkeeper
- Refresh Tab #1

You get it loading through to the dashboard, but without the user loaded (or something)
image

And then trying to use Logout from that state doesn't seem to kick you back to the login screen at this point, but refreshes through to a blank dashboard again.

jamshale:
@loneil This should be fixed now. I will test more but I think it's good.
I changed to new tab and logout flows to be more specific. I changed the login form to navigate home pages. Changed the paths on tenant (dashboard) and innkeeper (tenants). This helps distinguish between login and refresh (It wasn't doing this before and was setting the token for normal navigation.

Opened multiple tenant and innkeeper tabs and tried various logout/refresh/new tab scenarios.

loneil (approved):
Awesome to get this in. (wish we'd done it way back at the beginning of the app) 
I _think_ we're good here now. I can't see any weird behaviour, maybe there's some crazy edge cases I can't think of yet but might as well get it in wider use to see at this point.

Can merge assuming nothing wrong with rebase/branch update or anything.

jamshale:
If you have 2 tenant tabs open and logout of one then the other will logout on refresh. I think that behaviour is fine though and can't think of anyway around it. Would be better if it detected and logged out the other tab right away. That would be more of an enhancement imo.

I'm going to merge it.

PR #784, opened 2023-08-28
Make Tenant UI a default component in Traction chart

i5okie:
## Integrate Tenant-UI into Traction Helm Chart
- Integrate `tenant-ui` resources into Traction chart
- Add `ui.enabled` option to disable tenant-ui deployment when necessary
- Remove bespoke `tenant-ui` Helm chart


loneil:
@i5okie Regression tested all functionality and seems to be working on this PR deployment, except one thing.
When logging in through an OIDC user (like on the Innkeeper route, or a reservation) it calls through the Tenant UI backend, once logged in with OIDC, to the Traction URL and that URL seems to have the wrong config value.

This actually causes the pod to crash after a 404 in the Node API 🤯 , but that part's unelated to this (I must have missed a try/catch handler, will fix later).

It's this value in configmap.yaml for the UI
`  SERVER_TRACTION_URL: http://{{ include "acapy.fullname" . }}:{{ .Values.acapy.service.httpPort }}`

Resolves to 
image

wheareas it should be the same as the tenant proxy URL. Like in DEV below
image

So should be `https://pr-784-traction-tenant-proxy-dev.apps.silver.devops.gov.bc.ca` here

Possibly don't need both those config vars... but I can't remember if there was a good reason for the Vue frontend and the Node API to possibly have different values... probably can leave as is for now and just fix the SERVER_TRACTION_URL one

charts/traction/templates/ui/configmap.yaml
```diff
@@ -0,0 +1,31 @@
...
+  FRONTEND_TENANT_SHOW_WRITABLE_COMPONENTS: {{ .Values.ui.oidc.showWritableComponents | quote }}
+  FRONTEND_TENANT_PROXY_URL: https://{{ include "tenant_proxy.host" . }}
+  SERVER_OIDC_JWKS: {{ .Values.ui.oidc.jwksUri | quote }}
+  SERVER_OIDC_REALM: {{ .Values.ui.oidc.realm | quote }}
+  SERVER_OIDC_ROLE: {{ .Values.ui.oidc.roleName | quote }}
+  SERVER_SMTP_INNKEEPER_INBOX: {{ .Values.ui.smtp.innkeeperInbox | quote }}
+  SERVER_SMTP_PORT: {{ .Values.ui.smtp.port | quote }}
+  SERVER_SMTP_SENDER_ADDRESS: {{ .Values.ui.smtp.senderAddress | quote }}
+  SERVER_SMTP_SERVER: {{ .Values.ui.smtp.server | quote }}
+  SERVER_TRACTION_URL: http://{{ include "acapy.fullname" . }}:{{ .Values.acapy.service.httpPort }}
```
loneil, line 24: I think can also just be
`https://{{ include "tenant_proxy.host" . }}`
?

(see fuller comment in PR)

loneil (changes requested):
Just one issue with URL in node api

loneil (dismissed):
Think it's looking good now in this PR.

I can test out again when deployed to DEV to make sure all the values are good for that env too.

charts/traction/templates/ui/configmap.yaml
```diff
@@ -13,15 +13,15 @@ data:
...
+  FRONTEND_TENANT_PROXY_URL: https://{{ include "tenant_proxy.host" . }}
   SERVER_OIDC_JWKS: {{ .Values.ui.oidc.jwksUri | quote }}
   SERVER_OIDC_REALM: {{ .Values.ui.oidc.realm | quote }}
   SERVER_OIDC_ROLE: {{ .Values.ui.oidc.roleName | quote }}
   SERVER_SMTP_INNKEEPER_INBOX: {{ .Values.ui.smtp.innkeeperInbox | quote }}
   SERVER_SMTP_PORT: {{ .Values.ui.smtp.port | quote }}
   SERVER_SMTP_SENDER_ADDRESS: {{ .Values.ui.smtp.senderAddress | quote }}
   SERVER_SMTP_SERVER: {{ .Values.ui.smtp.server | quote }}
-  SERVER_TRACTION_URL: http://{{ include "tenant_proxy.fullname" . }}
+  SERVER_TRACTION_URL: https://{{ include "tenant_proxy.host" . }}
```
esune, line 24: @loneil is there a reason to have two environment variables set with the same value? Would we ever have the values be different (i.e.: can we have `SERVER_TRACTION_URL` route internally using the service name?), or should we collapse them into one environment variable?

loneil: Thinking about it I'd keep it as it is rather than changing anything. One is amongst the group of frontend configuration for the client side app, the other is for the Node API.
So while they're the same value (and likely to be the same value) I think having one shared configuration while the rest are split between the 2 apps would be more confusing.

As well, there could be cases to have them behave differently, as Ivan suggested earlier could have the API one be an internal route sometime. Also if we want to split the client side SPA to a different serving paradigm (like Caddy or something rather than being served by the same Node API) then they wouldn't have the same env vars.

loneil (approved):

PR #788, opened 2023-09-01
Self serve API key mgmt

loneil:
**Traction backend**
- Innkeeper plugin, add relevant TENANT endpoints for API key CRUD 
image
- All the same code paths for managing keys as the Innkeeper ones I created recently (though only for a specific Tenant obviously)
- Same as Innkeeper, keys are salt/hash, only returned in the POST result then not fetchable again
- Thoroughly tested you can only operate on keys for the token for the wallet you are operating on
- Innkeeper can still delete your keys and see records (though not password itself)
- You can manage keys an Innkeeper creates for you

**Tenant UI**
- Frontend for key CRUD, again same as shown Innkeeper one
- Added as as sidebar item and page, though I'm not convinced this is the right place for this in the UI.
The main pages and sidebar items are really for wallet functionality, maybe this belongs better in Settings or somewhere under the profile badge? We should UX review this along with lots of other stuff. But doing it here for now is the simplest.
- Need to add some unit tests, but will see if component organization gets reconfigured to move elsewhere first....
- Probably some code duplication resulting here since similar components as the Innkeeper are used, but different enough to not reuse.
image

### DIFF
--- a/plugins/traction_innkeeper/traction_innkeeper/v1_0/innkeeper/routes.py
+++ b/plugins/traction_innkeeper/traction_innkeeper/v1_0/innkeeper/routes.py
@@ -220,18 +220,18 @@ class TenantAuthenticationsApiRequestSchema(OpenAPISchema):
     )
 
     alias = fields.Str(
-        required=False,
-        description="Optional alias/label",
-        example="API key for sample line of buisness",
+        required=True,
+        description="Alias/label",
+        example="API key for sample line of business",
     )
 
 
 class TenantAuthenticationsApiResponseSchema(OpenAPISchema):
     """Response schema for api auth record."""
 
-    reservation_id = fields.Str(
+    tenant_authentication_api_id = fields.Str(
         required=True,
-        description="The reservation record identifier",
+        description="The API key record identifier",
         example=UUIDFour.EXAMPLE,
     )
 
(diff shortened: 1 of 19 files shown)


esune (approved):
Looks good to me 👍🏻 

Regarding code duplication: it looks like `DeleteApiKey`, `CreateApiKey` could, potentially, be the same component in innkeeper and tenant?

Regarding UX: I agree with the approach, a new navbar item is not really what we want, but I think we will need to revamp the UI/UX at a more general level and we might just move stuff around then rather than nesting it now and making our lives more complicated.

loneil:
> Looks good to me 👍🏻
> 
> Regarding code duplication: it looks like `DeleteApiKey`, `CreateApiKey` could, potentially, be the same component in innkeeper and tenant?
> 
> Regarding UX: I agree with the approach, a new navbar item is not really what we want, but I think we will need to revamp the UI/UX at a more general level and we might just move stuff around then rather than nesting it now and making our lives more complicated.

Yeah the thing is they have to call different endpoints whether you are the innkeeper or tenant, and the create one needs the different form with the tenant selector if you're the innkeeper. Could parameterize them with conditionals but still have some general app-wide code duplication reduction i'd like to look into for some other areas that have this same CRUD pattern.

esune:
I'm good with dealing with code duplication when we review the Tenant UI code in general at this point 👍🏻

PR #792, opened 2023-09-05
Prepare for 0.3.1 release

esune:


charts/traction/values.yaml
```diff
@@ -17,7 +17,7 @@ acapy:
     repository: ghcr.io/bcgov/traction-plugins-acapy
     pullPolicy: IfNotPresent
     pullSecrets: []
-    tag: ''
+    tag: '0.3.1'
```
i5okie, line 20: This should stay empty.

charts/traction/Chart.yaml
i5okie, line 1: Why are we jumping the minor versions here?
I think going from 0.1.0 to 0.2.0 was justified by the major change of merging `tenant-ui` into `traction` chart.
`version` is the Helm Chart version and should be separate from the app version.

Maybe `version` could be bumped to `0.2.1` since we're just updated the image tag?

charts/traction/README.md
```diff
@@ -1,6 +1,6 @@
 # Traction
 
-![version: 0.1.1](https://img.shields.io/badge/Version-0.1.0-informational?style=flat-square) ![Type: application](https://img.shields.io/badge/Type-application-informational?style=flat-square) ![AppVersion: 0.3.0](https://img.shields.io/badge/AppVersion-0.3.0-informational?style=flat-square)
+![version: 0.1.1](https://img.shields.io/badge/Version-0.1.0-informational?style=flat-square) ![Type: application](https://img.shields.io/badge/Type-application-informational?style=flat-square) ![AppVersion: 0.3.1](https://img.shields.io/badge/AppVersion-0.3.1-informational?style=flat-square)
```
i5okie, line 3: `![version: 0.1.1]` this should match the Helm chart version.

i5okie (changes requested):

charts/traction/Chart.yaml
esune: I didn't realize it had already been bumped, should I have looked at the line blame - will change to minor version bump.

charts/traction/values.yaml
esune: Ah, i see - I thought it would grab the value from here, but it picks up the chart value *unless* this is specified. Will revert.

loneil:
A couple places we could sync up the version labelling too. (Probably could have a more formal plan for versioning for the Tenant UI if it matters too)

- Tenant UI `package.json` files (there's 2 of them, API and Vue FE)
- I know it's gross and we should be pulling this from somewhere automatically, but **hard coded** in the display page `Traction.vue`

esune:
> A couple places we could sync up the version labelling too. (Probably could have a more formal plan for versioning for the Tenant UI if it matters too)
> 
>     * Tenant UI `package.json` files (there's 2 of them, API and Vue FE)
> 
>     * I know it's gross and we should be pulling this from somewhere automatically, but for now **hard coded** in the display page `Traction.vue`. Worth at least updating the hard-code when doing a release so if someone goes to the About page it can be used to know which release.

Updated, thanks for the tip. We really need to get this type of information consolidated, and figure out how versioning of each component affects the actual release: right now we have it matching the release, but technically each component (tenant UI, traction plugins) could be at different versions AND the release be a different version from them altogether since it "collects" them  at a certain point in time. Will give it some thought.

i5okie (approved):

Issue #775, opened 2023-08-22
Innkeeper UI changes for configurable reservation forms


loneil:
As an Innkeeper Tenant UI user, I want to see the details of a reservation record even if it's a dynamic form, so that I can inspect reservations before actioning them.
--

Right now the relational fields in a reservation work fine in the Innkeeper reservations and history tables.
Example:
![image](https://github.com/bcgov/traction/assets/17445138/194cb508-beff-41a4-b229-7dadf88457d7)

If the fields are dynamic and customizable need to know how to handle that in the table

AC:
-TBD

hiteshgh:
minimal scope/AC

loneil:
Easiest thing here
- Remove the reservation form columns that are not the 2 mandatory ones. SO just leave Tenant Name and Email
- Add the typical Row expander UI to display the whole Reservation, where the custom form fields would be

I would do at least the above before merging in the Reservation form ticket, should be quick to do this as a MVP.

In the future (or if a priority now, but do above first), could have the form definition config contain some metadata about "Show in Innkeeper Table" or something, and then dynamically add the columns the system wants shown at a glance.

cc @esune @popkinj

esune:
Resolved by #645

Issue #773, opened 2023-08-22
Reservation persistence changes to support customizable reservation form


loneil:
As a Traction operator I want the data model for Reservations to support extensible/customizable fields, so that a dynamic reservation form can be used
--

This would be required for proposed feature in https://github.com/bcgov/traction/issues/645
Innkeeper plugin would need to be changed to have the Reservation data model be non-relational for the reservation form.
However certain parts of a reservation (email, tenant name) are required as relational data for programmatic flows around creating tenants, so those could remain as relational fields while the rest is non.

AC
- The Innkeeper plugin persistence layer is changed so that the form fields in the reservation record are stored non-relationally (in a JSON blob or something). Or alternate solution for this is come up with.
- The email and tenant name fields do not change, they are needed as-is
- All existing reservation functionality still works as before.
- Some consideration or solution about what to do about existing reservation records that have relational fielded form data. Either need a transform or an understanding those fields' data will be lost or something.

loneil:
From discussions with @esune / @popkinj

esune:
We could have a database migration deal with the existing data, assuming we are using an ORM to manage entities and migrations.

PR #793, opened 2023-09-05
Custom reservation 645

popkinj:
## Custom Reservation Form.
Screenshot from 2023-09-05 12-26-35
Two form elements are mandatory:
1. Email
2. Name

All additional elements are optional and totally customizable..

Additional elements can be added by editing the [reservation.json](https://github.com/bcgov/traction/blob/custom-reservation-645/services/tenant-ui/frontend/public/forms/reservation.json) file.

The mandatory attributes (email & name) get entered into the database in traditional columns. All remaining attributes get packaged as a separate object and are stored as a JSON blob in the database.

Currently all additional data is stored as `context_data` and returned from the API. However, the UI for the Innkeeper isn't configured to use this yet. I will create a new ticket for this, as we haven't worked out how to handle it on the front end.
Screenshot from 2023-09-05 10-37-06

The tool used for generating custom forms is [jsonforms](https://jsonforms.io/). There are many options available if we choose to extend things in the future... Like more advanced form validation etc...

When this PR get's deployed, I will be interested to see how the reservation email behaves... as I wasn't able to test this locally.


loneil:
**Innkeeper**
We might have to action https://github.com/bcgov/traction/issues/775 as part of this before merging (or some hard coded workaround), otherwise it's lacking things on the Innkeeper side
image

**Form validation**
The validators are pre-triggered when you go into the form. Not sure if intended (other forms lazy validate on input/submit)
image

Can submit a invalid email format
image

loneil:
Everything on emails looking good **except** the Approval email that goes back to the reserver is missing their name (which makes sense as name may not be provided now):
image

If name is not a required field this may need to be tweaked to just have their email address or some generic "Applicant" text or something.

popkinj:
> **Innkeeper** We might have to action #775 as part of this before merging (or some hard coded workaround), otherwise it's lacking things on the Innkeeper side image
> 
> **Form validation** The validators are pre-triggered when you go into the form. Not sure if intended (other forms lazy validate on input/submit) image
> 
> Can submit a invalid email format image

- Can definitely add some email format validation.
- Glad to see a ticket for the Innkeeper tables. Not sure if it should block this PR seeing as nothing is broken. But we can definitely bump #775 to my next task.
- Yeah, I noticed that validation was kicking off on page load. If we want to go the lazy/load route, I can have a peak at that.

loneil:
> Glad to see a ticket for the Innkeeper tables. Not sure if it should block this PR seeing as nothing is broken. But we can definitely bump https://github.com/bcgov/traction/issues/775 to my next task.

I probably would, as an Innkeeper can't seem to see the prospective Tenant's requested Tenant Name which is needed for approval. Though the Tenant ID is a required field isn't it? So it should be showing up there even in this new paradigm?

For the other missing info, as a stopgap could add the expander to the Innkeeper table and remove all non-required fields from the table or something.

popkinj:
> > Glad to see a ticket for the Innkeeper tables. Not sure if it should block this PR seeing as nothing is broken. But we can definitely bump #775 to my next task.
> 
> I probably would, as an Innkeeper can't seem to see the prospective Tenant's requested Tenant Name which is needed for approval. Though the Tenant Name is a required field isn't it? So it should be showing up there even in this new paradigm?
> 
> For the other missing info, as a stopgap could add the expander to the Innkeeper table and remove all non-required fields from the table or something.

It looks like both the mail and table are using the `full_name` attribute...  This should be an easy switch to the mandatory name. We could even look for both if required.

loneil:
> > > Glad to see a ticket for the Innkeeper tables. Not sure if it should block this PR seeing as nothing is broken. But we can definitely bump #775 to my next task.
> > 
> > 
> > I probably would, as an Innkeeper can't seem to see the prospective Tenant's requested Tenant Name which is needed for approval. Though the Tenant Name is a required field isn't it? So it should be showing up there even in this new paradigm?
> > For the other missing info, as a stopgap could add the expander to the Innkeeper table and remove all non-required fields from the table or something.
> 
> It looks like both the mail and table are using the `full_name` attribute... This should be an easy switch to the mandatory name. We could even look for both if required.

Ahhhhh I see what's up here. It's been using the wrong field for a while now! Or forever?!

In `formattedReservations` do you mind changing 

`    tenant_name: msg.contact_name,`
To `msg.tenant_name` instead?

esune:
> 
>     * Yeah, I noticed that validation was kicking off on page load. If we want to go the lazy/load route, I can have a peak at that.

I think that would look better: if it is a big effort we could log it as a bug and deal with it after this has been merged. The other thing I am wondering is about the ordering of fields: I think it would make sense to display email and tenant name at the top, together, since they are the two fields that are always mandatory.


> 
> Not sure in the mail if it would want to say the Tenant's requested name as the greeting, or not... maybe a question for Emiliano.

We could use that, at least as a stop-gap. We may review the wording of the email in the future, and possibly make it so the template can be updated without rebuilding the whole app and allowing any `context_data` field to be used (potentially).

services/tenant-ui/frontend/public/forms/reservation.json
```diff
@@ -0,0 +1,52 @@
...
+      },
+      "phoneNumber": {
+        "type": "string"
+      },
+      "tenantName": {
+        "type": "string"
+      },
+      "tenantReason": {
+        "type": "string"
+      }
```
esune, line 19: I am not sure I love the idea of having the mandatory, non-removable fields (`emailAddress` and `tenantname`) as part of the dynamic payload: it may make sense to have these two in "traditional" input components instead, so that this file can be defined as empty if nothing more needs to be added.

loneil: Yeah as this part will eventually be user-config-defined, the mandatory fields might not really be appropriate to be defined there...
Or just have it be documentation that you must have those in there and it will fail if you don't 🤷 . We might already have it 422 on the endpoint if you don't supply them.

popkinj: Agreed. This is just the default and simplest config layout. We could definitely bust things apart as we see fit.

loneil:
Wondering if email valid should have any invalid text under the box. It pops an alert up so at least stops you but if there's a configuration field in the form definition that can do this?

image

popkinj:
> Wondering if email valid should have any invalid text under the box. It pops an alert up so at least stops you but if there's a configuration field in the form definition that can do this?
> 
> image

Yeah... I was aware of that.
Getting the jsonforms tool to properly verify the email was a really big pain. I just wrote a quick little regex to verify it. It's kinda a compromise.

loneil:
Biggest question on this now I think is from the final Acceptance Criteria on https://github.com/bcgov/traction/issues/773

> Some consideration or solution about what to do about existing reservation records that have relational fielded form data. Either need a transform or an understanding those fields' data will be lost or something.

We do have reservation data in Traction Dev and Test environments for integration work (I'm not sure how finalized any of it is?) so do those need to be preserved? I'm not sure what happens to the existing relational data contact fields on update? Is it just adding the `context_data` one and we'd need to manually remove the old fields? Or would they be lost?

services/tenant-ui/frontend/src/components/innkeeper/reservation/ReservationsHistory.vue
```diff
@@ -212,10 +161,7 @@ const formattedReservationHistory = computed(() =>
     state: msg.state,
     reservation_id: msg.reservation_id,
     contact_email: msg.contact_email,
-    contact_name: msg.contact_name,
-    contact_phone: msg.contact_phone,
     tenant_name: msg.contact_name,
```
loneil, line 164: Can do the same fix `tenant_name` instead of contact_name while we're at it on this table too.

services/tenant-ui/frontend/src/components/reservation/Reserve.vue
```diff
@@ -219,5 +173,42 @@ const handleSubmit = async (isFormValid: boolean) => {
     console.error(err);
     toast.error(`Failure making request: ${err}`);
   }
+
+  console.log('form is valid');
 };
 </script>
+<style scoped lang="scss">
```
loneil, line 180: Can we add a bit of top margin here between inputs so the form is spaced out as others are?

services/tenant-ui/frontend/src/components/innkeeper/reservation/ReservationsHistory.vue
popkinj: Done.

services/tenant-ui/frontend/src/components/reservation/Reserve.vue
popkinj: Sure thing. You mean like the previous reservation UI?

loneil: Right, just some space like other forms have (not sure we're entirely consistent with exact spacing accross the app) but they're generally spaced out more.
Ex:

image

esune:
> 
> First question is probably requirement of if we do need to know about existing reservations? Decision for @esune maybe?

It would be best to somehow migrate the data, if possible. I am not too concerned as we only have few clients right now, but keeping things working smoothly is definitely the preferred path.

services/tenant-ui/frontend/public/forms/reservation.json
esune: I would really prefer to have those fields defined (statically) elsewhere to limit the possibility of issues caused by human error or lack of attention. I realize it is not as straightforward as we're now mixing-up dynamic and non-dynamic fields, but I think that the desired "default" outcome would be to have reservations work even if the custom JSON is empty/missing.

services/tenant-ui/frontend/src/components/reservation/Reserve.vue
popkinj: Screenshot from 2023-09-08 12-03-31

popkinj:
Testing with an empty reservation.json file and with it completely missing. Both behaved as required by displaying the mandatory two fields.
Screenshot from 2023-09-12 10-39-47

services/tenant-ui/frontend/public/forms/reservation.json
popkinj: Done :)

esune:
> > First question is probably requirement of if we do need to know about existing reservations? Decision for @esune maybe?
> 
> It would be best to somehow migrate the data, if possible. I am not too concerned as we only have few clients right now, but keeping things working smoothly is definitely the preferred path.

@popkinj how will the deployment behave when upgrading from the current explicit colulmns to the JSON blobs?

Additionally, we may want to have the fields with validation errors be highlighted

popkinj:
> > > First question is probably requirement of if we do need to know about existing reservations? Decision for @esune maybe?
> > 
> > 
> > It would be best to somehow migrate the data, if possible. I am not too concerned as we only have few clients right now, but keeping things working smoothly is definitely the preferred path.
> 
> @popkinj how will the deployment behave when upgrading from the current explicit colulmns to the JSON blobs?
> 
> Additionally, we may want to have the fields with validation errors be highlighted

I don't really know how the db part works. I've never used this type of Python abstraction from the database before. My guess is we'll just have a new column that stores the jsonb data type. The old columns for phone number etc... will still be there but no longer filled.

As for the highlighting the mandatory fields... I'll have a look at what classes get added by the jsonforms library. I'm sure it's just a simple CSS thing.

popkinj:
Screenshot from 2023-09-13 09-44-52

popkinj:
> When going to see the reservation data, something like this is displayed:
> 
> ```
> {
> "state": "checked_in",
> "reservation_id": "893dae0f733d46c9ba4342264bb551b2",
> "contact_email": "email@mail.com",
> "tenant_name": "yayaya",
> "contact_name": undefined,
> "contact_phone": undefined,
> "tenant_reason": undefined,
> "context_data": {
> "phoneNumber": "123",
> "fullName": "ccc",
> "tenantReason": "ccc"
> },
> "created_at": "2023-09-14T17:45:47.085061Z",
> "created": "September 14 2023, 10:45:47 AM"
> }
> ```
> 
> I think the discrete attributes `contact_name`, `contact_phone` and `tenant_reason` should not be returned, since we removed them?

Yeah definitely.

popkinj:
> > When going to see the reservation data, something like this is displayed:
> > ```
> > {
> > "state": "checked_in",
> > "reservation_id": "893dae0f733d46c9ba4342264bb551b2",
> > "contact_email": "email@mail.com",
> > "tenant_name": "yayaya",
> > "contact_name": undefined,
> > "contact_phone": undefined,
> > "tenant_reason": undefined,
> > "context_data": {
> > "phoneNumber": "123",
> > "fullName": "ccc",
> > "tenantReason": "ccc"
> > },
> > "created_at": "2023-09-14T17:45:47.085061Z",
> > "created": "September 14 2023, 10:45:47 AM"
> > }
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > I think the discrete attributes `contact_name`, `contact_phone` and `tenant_reason` should not be returned, since we removed them?
> 
> Yeah definitely.

My thinking was to keep these for backwards compatibility.... Are you aware of any reason to keep them @esune and @loneil ?

loneil:
I don't think we'd be wanting to return those fields if they are not used any more. Therein the question about deleting them from the model, but as mentioned if we don't care about preserving existing Dev and Test reservations then they should be out of the model. (they should be out of the model **regardless**, the question was about a data fix to move them to new model)

Would recommend checking dev and test and at least someone manually pulling any necessary contact information into someone's notes or something (if anything needed).

popkinj:
> I don't think we'd be wanting to return those fields if they are not used any more. Therein the question about deleting them from the model, but as mentioned if we don't care about preserving existing Dev and Test reservations then they should be out of the model. (they should be out of the model **regardless**, the question was about a data fix to move them to new model)
> 
> Would recommend checking dev and test and at least someone manually pulling any necessary contact information into someone's notes or something (if anything needed).

Makes sense. I'll remove them from the model.
Looks like I'll also need to remove them from the tests.

esune:
Yep, let's just get rid of them. Luckily we do not have any need for those and we do not have a production instance that requires preserving existing values.

loneil:
@popkinj note changes to services/tenant-ui/frontend/src/types/acapyApi/acapyInterface.ts aren't intended to be made manually. That's the TS generation from the spec.

Can see https://github.com/bcgov/traction/tree/main/services/tenant-ui/frontend#generating-typescript-interfaces-and-types-from-the-aca-py-api

popkinj:
> @popkinj note changes to services/tenant-ui/frontend/src/types/acapyApi/acapyInterface.ts aren't intended to be made manually. That's the TS generation from the spec.
> 
> Can see https://github.com/bcgov/traction/tree/main/services/tenant-ui/frontend#generating-typescript-interfaces-and-types-from-the-aca-py-api
> 
> Removing things probably fine but may want to do the generator instructions to see.

Ah... K. Thanks.
I'll circle back to that after I finish my tests.

esune (approved):
Looks good to me 👍🏻

PR #805, opened 2023-09-07
Normalize tenant fields

jamshale:
Changed the models selialized values from `connected_to_endorsers` and `connect_to_endorsers` to `connect_to_endorser` for consistency and updated the ui.

Note: The attribute field used to change the serialization value will break if the python library `marshmallow` get upgraded to version ^3.0.0

### DIFF
--- a/plugins/traction_innkeeper/traction_innkeeper/v1_0/innkeeper/models.py
+++ b/plugins/traction_innkeeper/traction_innkeeper/v1_0/innkeeper/models.py
@@ -219,10 +219,11 @@ class ReservationRecordSchema(BaseRecordSchema):
         example=UUIDFour.EXAMPLE,
     )
 
-    connect_to_endorsers = fields.List(
+    connect_to_endorser = fields.List(
         fields.Dict(description="Endorser and ledger config", required=False),
         example=json.dumps(ENDORSER_LEDGER_CONFIG_EXAMPLE),
         required=False,
+        attribute="connect_to_endorsers"
     )
 
     create_public_did = fields.List(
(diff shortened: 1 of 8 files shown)


jamshale:
Going to test a bit more...

esune:
Since `marshmallow` is already beyond `3.0.0`, is there a different way to deal with this so that we will be able to update the dependency if/when needed?

jamshale:
The `attribute` parameter was changed to `data_key` in 3.0.0. It might be possible to have both parameters and not break anything for backwards/forwards compatibility though. I will have a look at that.

jamshale:
So I actually had a misunderstanding. The plugins are on marshmallow ^3.0.0. I got confused by another library in the poetry lock file that required ^2.15.2 

The `attribute` is an option for marshmallow ^3.0.0 and ^2.0.0. But `data_key` is only valid for ^3.0.0. So the changes i made are valid. `attribute` and `data_key` are actually opposites  of each other. 

These would all be equivalent.

Screenshot from 2023-09-08 13-41-42

Screenshot from 2023-09-08 13-42-19

Screenshot from 2023-09-08 13-43-05


https://stackoverflow.com/questions/56708349/what-is-the-difference-between-a-marshmallow-field-defined-with-data-key-and-o

esune:
Thank you @jamshale - @loneil were there other attributes that would require normalization, or are we good with these changes for the time being?

loneil:
There might be but I haven't looked all over the API or anything. Just had noticed these 2 while we were developing the auto-issuance stuff.

jamshale:
I looked for a little bit. at the api and didn't see anything. I'll have a more thorough look with this ticket. Would be nice to make sure and not have something annoying be found one day.

Also this doesn't deal with backwards compatibility but I was thinking because all the code is in the same repo we wouldn't have an old version of tenant-ui against the new innkeeeper-api and vice versa .

esune (approved):
👍🏻 looks good to me!

jamshale:
Don't see any other inconsistencies. I'm going to merge it as is.

PR #809, opened 2023-09-13
Split Helm Chart Releaser into bespoke workflow with manual trigger

i5okie:
The solution provided doesn't directly address our primary concern. However, by isolating the Helm Chart Releaser into its own workflow and adding a manual trigger, we can more easily troubleshoot the issue.

- Added `index.yaml` validation step to preemptively address potential conflicts in the future.

## Notes:
For a Helm chart releaser action to be successful, it's essential to update the `version` value in the `Chart.yaml` file, in addition to making other necessary changes to the Helm chart files.


cr.yaml
```diff
@@ -0,0 +1,4 @@
+owner: bcgobv
```
esune, line 1: Should be `bcgov`

esune (changes requested):

cr.yaml
i5okie: fixed

esune (approved):
👍🏻

PR #811, opened 2023-09-13
[Helm] Add mechanism to automatically restart pods on upgrade

i5okie:
- Added annotations to deployments with checksums for relevant configmaps and secrets


esune (changes requested):
Can we remove the `values-test.yaml` and `values-production.yaml` altogether since configurations for those environments are stored elsewhere?

Other than that, changes look good - it will be great to not have to kick the pods every time 😎

loneil:
> Can we remove the values-test.yaml and values-production.yaml altogether since configurations for those environments are stored elsewhere?

Maybe out of touch on this, but where are we storing those then?

esune:
> > Can we remove the values-test.yaml and values-production.yaml altogether since configurations for those environments are stored elsewhere?
> 
> Maybe out of touch on this, but where are we storing those then?

https://github.com/bcgov/trust-over-ip-configurations/tree/main/helm-values

i5okie:
> > > Can we remove the values-test.yaml and values-production.yaml altogether since configurations for those environments are stored elsewhere?
> > 
> > 
> > Maybe out of touch on this, but where are we storing those then?
> 
> https://github.com/bcgov/trust-over-ip-configurations/tree/main/helm-values

Should we also remove the [Promote Instance](https://github.com/bcgov/traction/blob/main/.github/workflows/on_push_tag.yaml) workflow?

esune (approved):
👍🏻

PR #813, opened 2023-09-13
Add inactivity timeout message

jamshale:
Adds a message to the login page when the gets logged out do to inactivity. 

Because the app is reloaded on logout it can't use state memory and needs to use local storage here. I only use one key for innkeeper and tenant because the key is erased on load so they shouldn't conflict with each other.

Added some very basic unit tests but haven't figured out how to mock local storage with return values yet. 

I think the info modal looks `ok` but am open for any suggestions how to make it look better. Thought of using a toast notification but didn't like it.

Screenshot from 2023-09-13 16-24-33


services/tenant-ui/frontend/src/plugins/i18n/locales/en.json
```diff
@@ -429,6 +429,8 @@
   },
   "session": {
     "countdown": "You are going to be logged out in { seconds } seconds...",
-    "header": "Session timing out"
+    "countdownHeader": "Session timing out",
+    "timeoutMessage": "Your session has timed out due to inactivity and you have been logged out. Please re-enter your login information.",
+    "timeoutHeader": " "
```
jamshale, line 434: Better way to have a blank header? Was having trouble figuring one out.

loneil: Hmm, haven't needed a blank field so nothing that I know.

loneil (approved):
Fantastic! Think the UX is as good as any, might get a whole-app UX review going at some point so could get opinions then. Agree it's better than the toast for sure (same style and format as the timing out warning which is best).

PR #817, opened 2023-09-15
Bringing back old attributes into the reservations model

popkinj:
Short term fix for API errors. 
Unable to duplicate error locally. I'm hoping test more thoroughly with the deployed PR.


popkinj:
Everything seem to work here. However, the DB obviously doesn't have any of the older data.
All the old attributes have be re-introduced to the reservation modal. However, I don't see them in the payload for the Innkeeper table. Is this a possible issue???
Screenshot from 2023-09-15 14-33-02

loneil:
> Everything seem to work here. However, the DB obviously doesn't have any of the older data. All the old attributes have be re-introduced to the reservation modal. However, I don't see them in the payload for the Innkeeper table. Is this a possible issue??? 
If they're empty, which they would be on a new reservation they wouldn't come back in the response, right? I think that's how it works.

popkinj:
> > Everything seem to work here. However, the DB obviously doesn't have any of the older data. All the old attributes have be re-introduced to the reservation modal. However, I don't see them in the payload for the Innkeeper table. Is this a possible issue???
> 
> If they're empty, which they would be on a new reservation they wouldn't come back in the response, right? I think that's how it works.

Ah.... That makes sense.

esune:
@loneil @popkinj if this is ready to go please merge, don't wait on me

loneil (approved):

PR #827, opened 2023-09-20
Form tweaks 818

popkinj:
Reservation form style tweaks and regex optimizations.


esune (approved):
Tweak looks good. Wondering if we can add an error message as well about what is failing with the validator (similar to the "this is a required field" message that comes out when a field is missing), but this might be for a separate task.

popkinj:
> Tweak looks good. Wondering if we can add an error message as well about what is failing with the validator (similar to the "this is a required field" message that comes out when a field is missing), but this might be for a separate task.

Do you mean the validation browser side or server side?

PR #829, opened 2023-09-21
Add other connection row advanced details

loneil:
Lets make debugging some issues (and getting details about established connections) a lot quicker. Call additional connection details when digging into a Connection row.
- Endpoints
- Metadata

This stuff could be hidden with the developer flag if we'd prefer once there's production usage.

Should help with
- Mediator testing
- Endorser connection checking
- Other service endpoint issues

Examples:

---
image

---
image

---
image

---
image

### DIFF
--- a/services/tenant-ui/frontend/src/assets/primevueOverrides/datatable.scss
+++ b/services/tenant-ui/frontend/src/assets/primevueOverrides/datatable.scss
@@ -79,3 +79,10 @@
     margin-right: 15px;
   }
 }
+
+// A specific divider used in expanded table rows
+.expand-divider {
+  border: 1px dashed grey;
+  width: 40px;
+  margin-inline-start: 0;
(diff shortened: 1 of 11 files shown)


jamshale:
I think more information is better as long as it doesn't cause to much confusion/clutter for end users. So I agree this make be too much in production.

Code looks good though :+1:

jamshale (approved):

loneil:
Considering whether **any** of those expanders that show JSON all over the app should be developer flag hidden in production. So agree it's something to think about hiding in the not too far future.

esune (approved):
I like it! 👍🏻 
Regarding hiding the functionality in production: I actually think it might be useful there more than anywhere to troubleshoot. Connections in particular is read/write in production as well to allow for OOB interactions, so this is great.

PR #838, opened 2023-09-25
Update aca py to 0.10.2

loneil:
Switch from rc0 to proper 0.10.2.

Have to set `typing-extensions = "4.5.0"` specifically to 4.5.0 since Pyddi seems to require 4.5>5, but aca-py requires < 4.6.

Change to commit the top level poetry lock in the plugins folder (was gitignored)


esune (approved):
👍🏻

PR #842, opened 2023-09-26
Custom form 774

popkinj:
Moving the custom form configuration from a stand alone JSON file to an environment variable.

Tested 
  - The local development environment
  - The docker environment
 
UNTESTED
  - The Helm configuration needs to be tested in the deployed PR environment.


esune:
Adding @i5okie as reviewer too. To test in the PR, you'll have to:
- Update the [pr-values.yml](https://github.com/bcgov/traction/blob/6f27861830048ea9e111905a697ae694cfcda3a4/deploy/traction/values-pr.yaml) to include a custom reservation form, different than the default one. I would suggest making teh default from "empty" and adding a description field maybe to the PR one?
- Update the tenant UI [deployment.yml](https://github.com/bcgov/traction/blob/6f27861830048ea9e111905a697ae694cfcda3a4/charts/traction/templates/ui/deployment.yaml) to mount the configmap in the right spot for the UI to pick it up.
I also don;t see changes to `docker-compose.yml` indicating the custom config file is being mounted at runtime - are we sure we are doing it?

popkinj:
> Adding @i5okie as reviewer too. To test in the PR, you'll have to:
> 
> * Update the [pr-values.yml](https://github.com/bcgov/traction/blob/6f27861830048ea9e111905a697ae694cfcda3a4/deploy/traction/values-pr.yaml) to include a custom reservation form, different than the default one. I would suggest making teh default from "empty" and adding a description field maybe to the PR one?
> * Update the tenant UI [deployment.yml](https://github.com/bcgov/traction/blob/6f27861830048ea9e111905a697ae694cfcda3a4/charts/traction/templates/ui/deployment.yaml) to mount the configmap in the right spot for the UI to pick it up.
>   I also don;t see changes to `docker-compose.yml` indicating the custom config file is being mounted at runtime - are we sure we are doing it?

Ah yes. I didn't add anything to the -pr and -development files. So I'm assuming it takes up the values.yaml by default????
How does one over ride the default config... Say you want no extra attributes. Would you just provide a blank string?

loneil:
> > Adding @i5okie as reviewer too. To test in the PR, you'll have to:
> > 
> > * Update the [pr-values.yml](https://github.com/bcgov/traction/blob/6f27861830048ea9e111905a697ae694cfcda3a4/deploy/traction/values-pr.yaml) to include a custom reservation form, different than the default one. I would suggest making teh default from "empty" and adding a description field maybe to the PR one?
> > * Update the tenant UI [deployment.yml](https://github.com/bcgov/traction/blob/6f27861830048ea9e111905a697ae694cfcda3a4/charts/traction/templates/ui/deployment.yaml) to mount the configmap in the right spot for the UI to pick it up.
> >   I also don;t see changes to `docker-compose.yml` indicating the custom config file is being mounted at runtime - are we sure we are doing it?
> 
> Ah yes. I didn't add anything to the -pr and -development files. So I'm assuming it takes up the values.yaml by default???? How does one over ride the default config... Say you want no extra attributes. Would you just provide a blank string?

That would depend on the json forms library right? Does it take just a blank string? An empty object?

popkinj:
@i5okie - I ended up going with an environment variable option... as it is much easier to integration in the development and docker environments. Having Helm swap files in and out may be a good enhancement for the future.

i5okie:
> @i5okie - I ended up going with an environment variable option... as it is much easier to integration in the development and docker environments. Having Helm swap files in and out may be a good enhancement for the future.

Can you add a mechanism to allow for both?
For example, if the environment variable for the form is empty, then look at the file.

i5okie:
> > > Adding @i5okie as reviewer too. To test in the PR, you'll have to:
> > > 
> > > * Update the [pr-values.yml](https://github.com/bcgov/traction/blob/6f27861830048ea9e111905a697ae694cfcda3a4/deploy/traction/values-pr.yaml) to include a custom reservation form, different than the default one. I would suggest making teh default from "empty" and adding a description field maybe to the PR one?
> > > * Update the tenant UI [deployment.yml](https://github.com/bcgov/traction/blob/6f27861830048ea9e111905a697ae694cfcda3a4/charts/traction/templates/ui/deployment.yaml) to mount the configmap in the right spot for the UI to pick it up.
> > >   I also don;t see changes to `docker-compose.yml` indicating the custom config file is being mounted at runtime - are we sure we are doing it?
> > 
> > 
> > Ah yes. I didn't add anything to the -pr and -development files. So I'm assuming it takes up the values.yaml by default???? How does one over ride the default config... Say you want no extra attributes. Would you just provide a blank string?
> 
> That would depend on the json forms library right? Does it take just a blank string? An empty object?

`values.yaml` are the defaults.
I believe if you try to supply something when overriding that is not in the values.yaml, it will be mad at you. So at very least, you should have an empty key in values. Otherwise, provide a default value if it's needed to work, and is expected to work for any and all expected end users of the helm chart.

esune:
If we can have both environment variable AND configmap it might be okay, as long as explicit priorities are set.

I strongly lean towards using a configmap/file mount for this though, I am not a fan of long JSON strings in environment variables: they're hard to read and troubleshoot with, when needed.

popkinj:
> If we can have both environment variable AND configmap it might be okay, as long as explicit priorities are set.
> 
> I strongly lean towards using a configmap/file mount for this though, I am not a fan of long JSON strings in environment variables: they're hard to read and troubleshoot with, when needed.

Yeah... I'm not super keen on it either. It just seemed like a good compromise.

Here's a silly question... Is it possible to have the config set in YAML and then have Helm pass it to the front end as a JSON object or string?

esune:
> > If we can have both environment variable AND configmap it might be okay, as long as explicit priorities are set.
> > I strongly lean towards using a configmap/file mount for this though, I am not a fan of long JSON strings in environment variables: they're hard to read and troubleshoot with, when needed.
> 
> Yeah... I'm not super keen on it either. It just seemed like a good compromise.
> 
> Here's a silly question... Is it possible to have the config set in YAML and then have Helm pass it to the front end as a JSON object or string?

@i5okie might have a better answer, but I know there is a `toJSON` function in Helm: https://helm.sh/docs/chart_template_guide/function_list/#type-conversion-functions

popkinj:
I'm going to close this PR for now as I'm making some substantial updates.

PR #854, opened 2023-10-03
Schema and credential definition improvements

jamshale:
Features:

- The main functionality of this feature is the ability to add multiple credential definitions for one schema. 
- Added the ability to copy schemas and credential definitions with validation. Added some messaging and error handling to make the features more clear.
- Added UX to connect schemas and cred defs.
- Added functionality to update the UI when schemas and cred defs get written to the ledger and are available.
- Added more acapy typing
- Created some reusable components that could be utilized in some other places.
- Refactored some code.

Notes:

- Changed the file structure in issuance directory to just have a folder for each parent component. Felt there was beginning to be too many sub directories.
- For the imports I felt like it was beginning to have too many categories and changed it so there is a section for libraries and src code.
- There seems to be problems with pagination in general when changing the number of returned rows. I think this should be looked at in another ticket.
- When the selected schema or cred def is active and the user navigates to the page it won't change to the correct page. Only goes to the first page. This might be a bit tricky to implement. I think it should be done in another ticket when it becomes more important.
- Would like to add some more unit tests but created the right mocks to test components using the governance store and added a few basic tests.


jamshale:
Going to bring back the copy/clone schema feature and make it more clear.

swcurran:
### Case 1: Reference a Schema

As an issuer, I want to be able entire the ID of a schema that already exists on the ledger, and put it in my secure storage such that I can create a credential definition based on the schema so that I can be an issuer of that type of schema.

Notes — the schema must exist on the ledger, I don’t have to be the publisher of the schema.

Use Case: A common “Business License” schema is published on the ledger, and all the Muncipalities in BC want to issue their business licenses using that common schema.

### Case 2: Version a Schema

As an issuer, I want to make a copy of an existing schema in my storage, edit the attributes (add, remove and modify) and save it as a new schema with (at least) a new version, so that I can create a new credential definition so that I can issue credentials that have new data elements.

Notes — the process is the same as creating a new schema, but with the existing schema’s name and attributes, with the version blank. The schema name and attributes can be changed as needed.

Use Case: The business has decided that they need to add an “expiry_date” field in the credentials being issued, with the rest of the attributes remaining the same.  It could be created from scratch, but easier to copy, edit and save.

jamshale:
@swcurran Sounds good. That makes total sense to me now.

services/tenant-ui/frontend/src/components/common/ValidatedField.vue
```diff
@@ -0,0 +1,84 @@
+<template>
```
loneil, line 1: Yeah this is perfect, we've needed this for a **long** time as the validation syntax is so verbose all over forms all over the app. 
We should expand to use this everywhere we can (and might need additional ones for other form controls) as we go through other components

The one thing in my mind on this also, is if we expand this common component to be used, do you think it's written in such a way that we could swap out the validation library if ever needed?
**Not saying we need to do this**, but Primevue when we first built this app had Vuelidate in it's documentation as the "suggested" vue validation framework, and some point recently they've changed it to VeeValidate

https://primevue.org/inputswitch/#veevalidate

services/tenant-ui/frontend/src/types/acapyApi/acapyInterface.ts
```diff
@@ -3378,7 +3378,7 @@ export interface SchemaStorageRecord {
   /** Schema identifier */
   ledger_id?: string;
   /** (Indy) schema */
-  schema?: object;
+  schema: Schema;
```
loneil, line 3381: This might possibly get overwritten on a subsequent generation of this file...
https://github.com/bcgov/traction/tree/main/services/tenant-ui/frontend/src/types#aca-py-api-types

Unless it's due to a change in the actual endpoint swagger definition. Maybe it **should** be updated if it's in a plugin to be represented this way?

services/tenant-ui/frontend/src/helpers/formatters.ts
```diff
@@ -0,0 +1,42 @@
...
+export interface FormattedSchema extends SchemaStorageRecord {
+  created: string;
+  credentialDefinitions: CredentialDefinition[];
+}
+
+export interface FormattedCredDef extends CredDefStorageRecord {
+  created: string;
+}
+
+export const formattSchemaList = (
```
loneil, line 18: 2 `t`s in these intentional?

services/tenant-ui/frontend/src/helpers/formatters.ts
```diff
@@ -0,0 +1,42 @@
+import {
```
loneil, line 1: Yeah I like this, can start moving the other table formatters in here.
Think it's intended on specifically containing just the object > datatable row formatters? Or other things too.
Could specify the file name as tableFormatters or something then maybe 🤷

services/tenant-ui/frontend/src/components/common/ValidatedField.vue
jamshale: This component would need to be changed obliviously but the best part is it only has to be change in one place. Then the validation rules for each form would need to be changed as well which would be quite a bit of work. The nice thing is though the all the form components using it would become obvious because of typescript. Once you changed the type to veevalidate.

There is other places this type of thing can be done but I figure we can just do them when we see them and when we have time.

services/tenant-ui/frontend/src/types/acapyApi/acapyInterface.ts
jamshale: Oh. I should have read the readme. I was wondering where this file came from.
Hmm. I'll need to think about it. There's probably some work around but I think the definition should definitely have the correct interface here instead of a generic object.

services/tenant-ui/frontend/src/helpers/formatters.ts
jamshale: No. Spelling mistake.

services/tenant-ui/frontend/src/components/common/ValidatedField.vue
loneil: Yeah I think it's knock them down as we change other related things.

Ok, validation library thing noted, as I said not thinking we should change but yeah it's good it could be more contained if we did. Mostly annoyed that the Primevue docs changed supported at some point.

services/tenant-ui/frontend/src/types/acapyApi/acapyInterface.ts
loneil: Could override/extend it in the other Types file maybe (or locally where it's used) until we check if the Aca-Py Swagger type can be improved.

loneil (dismissed):
Great stuff. Improvements to Schema and Cred Def pages working well from my testing (can create, copy, import other Tenants', make multiple defs, issue).
Other improvements looking good.

Minor note, just the changed modals have more padding, I don't really care which way (the old UX designs were specific about lining up/the lower padding ones) but we should try and make consistent, so if anyone reading does care about that, could put input on which one looks better

image


Approving, but just that one generated acapy Types file thing probably needs a tweak, or kick the can down the line to adjust later.

jamshale:
Great thanks. I'll fix up these few things before merging.

services/tenant-ui/frontend/src/types/acapyApi/acapyInterface.ts
jamshale: I changed this so there is another type in the types index file that extends the type with the correct interface. May need to do this for other types at some point. It's a little confusing but much safer.

jamshale:
> Minor note, just the changed modals have more padding, I don't really care which way (the old UX designs were specific about lining up/the lower padding ones) but we should try and make consistent, so if anyone reading does care about that, could put input on which one looks better

Oddly this only happened at exactly 400px. I changed the min width of this form and the padding went away. Must be something with the global library or some other css rule. Didn't dive into it much.

PR #870, opened 2023-10-15
fix: docker images

jleach:
To get "docker compose up" working I had to make a few small changes to the Dockerfile(s):

- I added `gcc` and `python3-devel` to `services/endorser/Dockerfile` because `pip3` failed to install the requirements because `gcc` was missing.

- I don't think any images for are stored in DockerHub so in `services/aca-py/Dockerfile.acapy` I sourced the image as follows: `FROM ghcr.io/bcgov/traction-plugins-acapy:sha-10e54d4`. These probably need proper image tags with versions or latest.

### DIFF
--- a/services/aca-py/Dockerfile.acapy
+++ b/services/aca-py/Dockerfile.acapy
@@ -1,4 +1,4 @@
-FROM traction:plugins-acapy
+FROM ghcr.io/bcgov/traction-plugins-acapy:sha-10e54d4
 
 USER root
 
--- a/services/endorser/Dockerfile
+++ b/services/endorser/Dockerfile
@@ -3,6 +3,10 @@ FROM python:3.10-slim-buster
 WORKDIR /app
 ENV ENDORSER_API_PORT=5000
 COPY requirements.txt requirements.txt
+
+RUN apt-get update && \ 
+  apt-get install gcc python3-dev -y
+
 RUN pip3 install -r requirements.txt
 
 COPY . .


loneil:
@jleach what's your computer setup? OS? 
To my knowledge no one else gets failures to build/start the endorser in the Docker Compose environment setup.
Are you just running `docker compose build` then `docker compose up` in the `scripts` folder? Or independently trying to start any of these dockerfiles?

jleach:
@loneil I run `docker compose up` which does a build on first run if needed. Yes, from `scripts`. Milage may varry based on when you built your images. For example, you have `pull_policy: missing` as your default so you call could be running very old images unless your recently rebuilt.

I started out from a clean env - no images at all AFAIK since this was my first run at starting traction.

jleach:
@loneil I tried with `docker compose build` on a different machine and it started up just fine. Will test a bit more before closing this issue.

esune:
I do get errors when trying to start Traction up from scratch related to the image built from `services/aca-py/Dockerfile.acapy`. So far, I have been able to work around them by following the "manual" build instructions [here](https://github.com/bcgov/traction/tree/main/scripts#build-images-directly). I believe the image is supposed to be built if missing, based on the service having both an image and build definition, but that doesn't seem to work well/reliably so it might have to change to having an explicit build command to start things up? I do not think that referencing a published image is the solution, though, as this would not pick-up changes made in the local environment on a rebuild.

Haven't really had issues with the endorser image build (at least yet): interested to know which OS this was triggered on, and whether it can maybe be fixed by pinning some dependencies.

esune:
Closing as this has become stale and no further feedback was provided, feel free to reopen if new information is available 🙂

PR #871, opened 2023-10-16
Handle real-world Endorser cases in Tenant UI

loneil:
Adjust all the Tenant UI profile ("make issuer") flow to handle asynchronous Endorser approvals.

So don't assume things can be a one-button action to connect to an endorser and publish a DID.
Break things back to the 2-step action, with additional handling for Endorser connection state and a bunch of other things.

However, **also** support situations where the Endorser IS set to auto-everything, like with our Sandbox environment etc.
To do this, allow a Tenant UI setting we'll call "Quick Connect"
` "quickConnectEndorserName": "bcovrin-test-endorser",` for example.
This is a single string right now, but could be an array, though it's likely just for single-ledger-setup cases right now, can adjust down the line.
If that's set then it waits for an endorser connection to be active then automatically moves on down to the register DID steps.

I need to catch up on some unit test stuff for this, but will put some time in on a different PR after finishing some other stuff.


loneil:
> Would it be possible to manage the `auto` endorser scenario without having to add an extra setting, that opens up the opportunity to mismatched configurations? Or is there something that is preventing us from having the "auto" scenario work the same as the "standard" pattern?

The Auto case could work the same as standard sure, then you can just manually refresh and manually do the DID step. Just figured for Sandbox and stuff it's helpful to have the Quick Connect concept.

esune:
Gotcha. I think I am fine as long as the logic 'still works" regardless of how the system is set-up (except for quick connect on non-auto endorsers, of course). My point was more about limiting the "required" configurations in favour of a more resilient handling of the flow.

jamshale (commented):
Code looks good :+1: 
Just curious about @esune 's comment as well. Don't want configurations to grow and grow if they aren't actually required. But I don't know the details enough to have a good opinion.

loneil:
> Gotcha. I think I am fine as long as the logic 'still works" regardless of how the system is set-up (except for quick connect on non-auto endorsers, of course). My point was more about limiting the "required" configurations in favour of a more resilient handling of the flow.

Yeah I did try "missing" or "wrong" config. And that stuff still all works.
- So if you have an auto-endorser, but don't mark it as quick you just have to hit buttons yourself
- If you have a non-auto, but DO mark it as quick. It still just sits in request state the same as if you didn't mark it (after waiting a bit for the auto accept that doesn't come back)

esune:
I am wondering if this should then be an "undocumented" option that we can use for development/sandboxwhile avoiding others to implement it... 🤔

loneil:
Yeah it's just a Tenant UI config val, and it defaults to blank so if no one does anything with it, it will just act as if it didn't exist anyways. In our helm vals I documented it (because we will set it), but again the default helm val is blank.
So another Traction operator doesn't have to do anything with it and can ignore it completely.

jamshale (approved):

PR #875, opened 2023-10-17
DO NOT MERGE - Per Tenant Logging Testing

shaangill025:


esune:
@shaangill025 the permission error we are seeing when creating the log file in OpenShift could be resolved by having the logs go into `/home/aries/log` rather than in `/home/aries`, as that folder should have the right privileges set for the user to write to it.
We probably want to have this applied as default folder in ACA-Py, rather than leaving it to the user.

shaangill025:
@esune I made the change in the following commit to allow `--log-file` to be applied to the logging handler if already created from config. So `--log-file ./log/acapy.log` would work with new pre-release image. Earlier the path specified in the config [`ini`] file could not be override. Regarding putting `./log/..` as the default path in ACA-Py, I am not sure if this is the way to go as this would work within Traction env but not if ACA-Py [multitenant] is stood up by itself. Currently the hang up is with the caplog issue with unit tests as I had mentioned in my updates today.
https://github.com/hyperledger/aries-cloudagent-python/pull/2550/commits/c46c4d0880b9d77e26521e9ea3d4502976c7c6a7

esune:
> @esune I made the change in the following commit to allow `--log-file` to be applied to the logging handler if already created from config. So `--log-file ./log/acapy.log` would work with new pre-release image. Earlier the path specified in the config [`ini`] file could not be override. Regarding putting `./log/..` as the default path in ACA-Py, I am not sure if this is the way to go as this would work within Traction env but not if ACA-Py [multitenant] is stood up by itself. Currently the hang up is with the caplog issue with unit tests as I had mentioned in my updates today. [hyperledger/aries-cloudagent-python@c46c4d0](https://github.com/hyperledger/aries-cloudagent-python/commit/c46c4d0880b9d77e26521e9ea3d4502976c7c6a7)

Thank you @shaangill025, i think I follow now. Looping in @WadeBarnes for potential input/insight on the operations end.

WadeBarnes:
For anyone using the ACA-Py images the `./log` folder is where logs are expected to be written if stored to disk.

swcurran:
@WadeBarnes — are you now happy with the implementation in ACA-Py?  As in approving the PR?

@shaangill025 — two of the test Actions are failing in the ACA-Py PR. Could you please investigate and address them?  We’ve had some trouble with the JSON-LD tests failing to resolve a URL that could be the cause of the integration test errors (or not…), but that would not be the issue with the other failing action.

WadeBarnes:
> @WadeBarnes — are you now happy with the implementation in ACA-Py? As in approving the PR?

I've been letting @esune lead on the review of the Per Tenant Logging.  I'm just commenting on the default location used for ACA-Py when it logs to disk.  The images setup a `./log` (more precisely a `~/log`) folder and relevant permissions for this purpose, but the defaults in ACA-Py seem to log to files in the `.` (more precisely the `~`) folder.

WadeBarnes:
The pods associated with this PR are still in a crash loop.
cc @shaangill025, @esune, @i5okie

PR #885, opened 2023-10-18
OOB invitations

loneil:
Allow choice between OOB connection type and connections/1.0 for creating an invitation.
Supports simple invitation protocol, can consider expanding to more complex handshake procotol stuff in the future but that would be best with a URL shortening plan (QR code definitely could have issues)

### DIFF
--- a/services/tenant-ui/frontend/src/components/connections/createConnection/CreateConnectionForm.vue
+++ b/services/tenant-ui/frontend/src/components/connections/createConnection/CreateConnectionForm.vue
@@ -53,7 +53,7 @@
     </div>
 
     <!-- Alias -->
-    <div class="field w-full">
+    <div class="field w-full mt-3">
       <label for="alias" :class="{ 'p-error': v$.alias.$invalid && submitted }">
         {{ $t('connect.invitation.alias') }}
       </label>
--- a/services/tenant-ui/frontend/src/components/connections/createConnection/CreateConnectionForm.vue
+++ b/services/tenant-ui/frontend/src/components/connections/createConnection/CreateConnectionForm.vue
@@ -1,5 +1,57 @@
 <template>
   <form @submit.prevent="handleSubmit(!v$.$invalid)">
+    <div v-if="!invitation_url">
+      <!-- Invitation type -->
+      <div class="mb-3">
+        <span>{{ $t('connect.invitation.type') }}</span>
+        <div class="my-2">
+          <RadioButton
+            v-model="isOob"
+            input-id="type1"
+            name="type"
+            :value="false"
+          />
+          <label for="type1" class="ml-2">{{
+            $t('connect.invitation.typeConnections')
+          }}</label>
+        </div>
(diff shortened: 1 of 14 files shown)


jamshale:
The failing tests were probably a bit to fragile. Changing the table columns would make them fail.

loneil:
> The failing tests were probably a bit to fragile. Changing the table columns would make them fail.
> 
> Could just remove the failing test or update it with the expected column value.

@jamshale yeah sorry had that fixed but got sidetracked with other PRs and hadn't requests a review on this one yet.
Should be good to go to take a look now, thanks.

services/tenant-ui/frontend/src/components/connections/createConnection/CreateConnectionForm.vue
```diff
@@ -1,7 +1,59 @@
...
+            v-model="isOob"
+            input-id="type1"
+            name="type"
+            :value="false"
+          />
+          <label for="type1" class="ml-2">{{
+            $t('connect.invitation.typeConnections')
+          }}</label>
+        </div>
+        <div>
```
jamshale, line 18: This extra div needed? The other group doesn't have one. 
It's only a tiny bit of code but these 2 elements consisting of title+radio+label+radio+radio look like they could be the same component.

services/tenant-ui/frontend/test/components/connections/Invitations.test.ts
```diff
@@ -20,16 +20,17 @@ describe('Invitations', () => {
...
 
   test('table body is rendered with expected values', async () => {
     const wrapper = mountInvitations();
-    const expectedTexts = ['', 'test.alias', 'once'];
+    const expectedTexts = ['', 'test.alias', 'once', 'connections/1.0'];
 
     // td is an expected text or valid date
     wrapper.findAll('tbody td').forEach((td) => {
       const text = td.text();
+      console.log(text);
```
jamshale, line 33: Probably left over from checking the test html.

jamshale (approved):

services/tenant-ui/frontend/src/components/connections/createConnection/CreateConnectionForm.vue
loneil: The divs around the Radio + Label elements? Without those things end up in-line
image

I might be misunderstanding

services/tenant-ui/frontend/test/components/connections/Invitations.test.ts
loneil: Thx! Fixed

services/tenant-ui/frontend/src/components/connections/createConnection/CreateConnectionForm.vue
jamshale: Oh, I see. Just the top one needed a margin.

jamshale (approved):

PR #886, opened 2023-10-20
Banner 806

popkinj:
Custom banner that can be configured for each environment. It is persistent ie. the user can not dismiss.

## Not logged in

**Info**
Screenshot from 2023-10-18 11-16-16

**Alert**
Screenshot from 2023-10-18 11-18-47

**Error or Serious**
Screenshot from 2023-10-18 11-17-32

**Turned off**
Screenshot from 2023-10-18 11-19-33

All the same styles apply to when logged in... But located in a different spot. Not totally sure about this yet.... Open to suggestions.
Screenshot from 2023-10-20 12-04-41

### DIFF
--- a/charts/traction/templates/ui/configmap.yaml
+++ b/charts/traction/templates/ui/configmap.yaml
@@ -35,6 +35,4 @@ data:
   UX_INFO_BANNER_MESSAGE: {{ .Values.ui.ux.infoBanner.message | quote }}
   UX_INFO_BANNER_MESSAGE_LEVEL: {{ .Values.ui.ux.infoBanner.messageLevel | quote }}
   UX_INFO_BANNER_SHOW_MESSAGE: {{ .Values.ui.ux.infoBanner.showMessage | quote }}
-  UX_INFO_BANNER_ENVIRONMENT_NAME: {{ .Values.ui.ux.infoBanner.environmentName | quote }}
-  UX_INFO_BANNER_SHOW_ENVIRONMENT_NAME: {{ .Values.ui.ux.infoBanner.showEnvironmentName | quote }}
 {{- end -}}
--- a/charts/traction/templates/ui/configmap.yaml
+++ b/charts/traction/templates/ui/configmap.yaml
@@ -32,4 +32,9 @@ data:
   UX_COVERIMAGE_COPYRIGHT: {{ .Values.ui.ux.coverImageCopyright | quote }}
   UX_OWNER: {{ .Values.ui.ux.owner | quote }}
   UX_SIDEBAR_TITLE: {{ .Values.ui.ux.sidebarTitle | quote }}
+  UX_INFO_BANNER_MESSAGE: {{ .Values.ui.ux.infoBanner.message | quote }}
+  UX_INFO_BANNER_MESSAGE_LEVEL: {{ .Values.ui.ux.infoBanner.messageLevel | quote }}
+  UX_INFO_BANNER_SHOW_MESSAGE: {{ .Values.ui.ux.infoBanner.showMessage | quote }}
+  UX_INFO_BANNER_ENVIRONMENT_NAME: {{ .Values.ui.ux.infoBanner.environmentName | quote }}
+  UX_INFO_BANNER_SHOW_ENVIRONMENT_NAME: {{ .Values.ui.ux.infoBanner.showEnvironmentName | quote }}
 {{- end -}}
--- a/services/tenant-ui/config/custom-environment-variables.json
+++ b/services/tenant-ui/config/custom-environment-variables.json
@@ -33,9 +33,7 @@
       "infoBanner": {
         "message": "UX_INFO_BANNER_MESSAGE",
         "messageLevel": "UX_INFO_BANNER_MESSAGE_LEVEL",
-        "showMessage": "UX_INFO_BANNER_SHOW_MESSAGE",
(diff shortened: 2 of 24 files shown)


loneil:
Don't really like the look of it in the header there in the last screenshot personally. Was thinking more like a topline banner above all the usual UX
image

loneil (changes requested):
Seeing the banner with an empty message when the fields are turned off

image

popkinj:
> Seeing the banner with an empty message when the fields are turned off
> 
> image

Nice grab. By turned off do you mean the showMessage boolean?

loneil:
> > Seeing the banner with an empty message when the fields are turned off
> 
> Nice grab. By turned off do you mean the showMessage boolean?



Not sure, whatever is set for values here in this PR has blank for text and false for showMessage and it's still showing. 

image

popkinj:
> > > Seeing the banner with an empty message when the fields are turned off
> > 
> > 
> > Nice grab. By turned off do you mean the showMessage boolean?
> 
> Not sure, whatever is set for values here in this PR has blank for text and false for showMessage and it's still showing.
> 
> image

ok... Thanks for the screenshot. That's definitely not the intended behaviour. I'll have a look.

esune:
> Don't really like the look of it in the header there in the last screenshot personally. Was thinking more like a topline banner above all the usual UX

I agree with Lucas: the banner looks good on the sign-in page as an alert, but it isn't ideal when inside the Tenant UI.
@popkinj can we adjust it to look more similar to the screenshot/recommendation from Lucas? If it has to be that way on the m ain sign-in screen for consistency it's fine

popkinj:
> > Don't really like the look of it in the header there in the last screenshot personally. Was thinking more like a topline banner above all the usual UX
> 
> I agree with Lucas: the banner looks good on the sign-in page as an alert, but it isn't ideal when inside the Tenant UI. @popkinj can we adjust it to look more similar to the screenshot/recommendation from Lucas? If it has to be that way on the m ain sign-in screen for consistency it's fine

Sure. This may just be a placement thing with a couple css tweaks. Will play around a bit.

popkinj:
Screenshot from 2023-10-23 15-16-28

popkinj:
Screenshot from 2023-10-23 15-19-14

popkinj:
Tested a bunch of configurations and things are looking really good now.

charts/traction/templates/ui/configmap.yaml
```diff
@@ -32,4 +32,9 @@ data:
   UX_COVERIMAGE_COPYRIGHT: {{ .Values.ui.ux.coverImageCopyright | quote }}
   UX_OWNER: {{ .Values.ui.ux.owner | quote }}
   UX_SIDEBAR_TITLE: {{ .Values.ui.ux.sidebarTitle | quote }}
+  UX_INFO_BANNER_MESSAGE: {{ .Values.ui.ux.infoBanner.message | quote }}
+  UX_INFO_BANNER_MESSAGE_LEVEL: {{ .Values.ui.ux.infoBanner.messageLevel | quote }}
+  UX_INFO_BANNER_SHOW_MESSAGE: {{ .Values.ui.ux.infoBanner.showMessage | quote }}
+  UX_INFO_BANNER_ENVIRONMENT_NAME: {{ .Values.ui.ux.infoBanner.environmentName | quote }}
+  UX_INFO_BANNER_SHOW_ENVIRONMENT_NAME: {{ .Values.ui.ux.infoBanner.showEnvironmentName | quote }}
```
loneil, line 39: As far as I can tell these are unused right?

services/tenant-ui/config/custom-environment-variables.json
```diff
@@ -29,6 +29,13 @@
         "linkTitle": "UX_ABOUT_LINK_TITLE",
         "linkText": "UX_ABOUT_LINK_TEXT",
         "imageUrl": "UX_ABOUT_IMAGE_URL"
+      },
+      "infoBanner": {
+        "message": "UX_INFO_BANNER_MESSAGE",
+        "messageLevel": "UX_INFO_BANNER_MESSAGE_LEVEL",
+        "showMessage": "UX_INFO_BANNER_SHOW_MESSAGE",
+        "environmentName": "UX_INFO_BANNER_ENVIRONMENT_NAME",
+        "showEnvironmentName": "UX_INFO_BANNER_SHOW_ENVIRONMENT_NAME"
```
loneil, line 38: Can't see these 2 referenced in the code.

loneil (changes requested):
Think we should remove unused config

charts/traction/templates/ui/configmap.yaml
popkinj: These are placeholders for one of the [blocked issues](https://github.com/bcgov/DITP-DevOps/issues/125).

popkinj: I can take them out of this branch if you want.

loneil: But for that issue it's just setting the `message` field to "Development Instance" etc
Don't need 2 different config settings when you can just set `message`. Unless it's intended to be 2 different UIs or something, but that's not supported in the view so yeah I'd take it out until something like that is needed. Can just use `message` for everything.

popkinj: Removed.

loneil (approved):

PR #895, opened 2023-10-25
Accept OOB invite

loneil:
Quick change for a better UX to pick between accept an OOB or Connections/1.0 invitation type

image

image

### DIFF
--- a/services/tenant-ui/frontend/src/components/connections/acceptInvitation/AcceptInviteForm.vue
+++ b/services/tenant-ui/frontend/src/components/connections/acceptInvitation/AcceptInviteForm.vue
@@ -1,5 +1,32 @@
 <template>
   <form v-if="urlEntryStep" @submit.prevent="handleSubmit(!v$.$invalid)">
+    <!-- Invitation type -->
+    <div class="mb-3">
+      <span>{{ $t('connect.invitation.type') }}</span>
+      <div class="my-2">
+        <RadioButton
+          v-model="isOob"
+          input-id="type1"
+          name="type"
+          :value="false"
+        />
+        <label for="type1" class="ml-2">{{
+          $t('connect.invitation.typeConnections')
+        }}</label>
+      </div>
+      <div>
+        <RadioButton
+          v-model="isOob"
+          input-id="type2"
+          name="type"
+          :value="true"
+        />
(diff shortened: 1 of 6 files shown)


services/tenant-ui/frontend/src/components/connections/acceptInvitation/AcceptInviteForm.vue
```diff
@@ -1,5 +1,32 @@
 <template>
   <form v-if="urlEntryStep" @submit.prevent="handleSubmit(!v$.$invalid)">
+    <!-- Invitation type -->
+    <div class="mb-3">
+      <span>{{ $t('connect.invitation.type') }}</span>
+      <div class="my-2">
```
jamshale, line 6: I do think a div containing a radio button and a label would be a good nested component. Just DRY stuff. This is small and not a big deal but i'm just going to try and point stuff out no matter how small.

jamshale (approved):

services/tenant-ui/frontend/src/components/connections/acceptInvitation/AcceptInviteForm.vue
jamshale: Ya it does get less beneficial if your passing a similar amount of props. This is a minimal situation where you would get any benefit, and it's probably not worth it.

I don't think you would need to use the name or input-id props. You would still need to send in a model, initial value and label value. Still if this was done in many places it would be beneficial imo, if nothing else it will just make the template section a bit cleaner and say you wanted to change the left margin for all of them you'd only need to do it in one place.

jamshale: Also you can define the child component with default values for any radio button props that are used. Then in some situations you should only need to have a model and label prop passed in.

If you find you need to use a radio button prop in the future you can declare a prop with the same default value and none of the existing components would be affected.

loneil: Yeah I guess one of the points of wrapping is also to get standardization, so if we'd be finding ourselves needing to override a bunch of props in a bunch of different places that would be a bad sign we're not keeping consistent UX as well.

PR #901, opened 2023-10-31
innkeeper: soft delete tenant

gurjmatharu:
Added soft delete
Note: we will need to create a follow up issue for hard delete.
- added delete confirmation thingamabober
  -  869-PR-1
- added query params to `/innkeeper/tenants` which defaults to `active`.  If `all` is supplied as a param it will return all tenants. 
  - 869-PR-2 
- added soft delete endpoint
  - 869-PR-3

### DIFF
--- a/services/tenant-ui/frontend/src/components/innkeeper/tenants/Tenants.vue
+++ b/services/tenant-ui/frontend/src/components/innkeeper/tenants/Tenants.vue
@@ -33,7 +33,10 @@
       <Column :expander="true" header-style="width: 3rem" />
       <Column :sortable="false" :header="$t('common.actions')">
         <template #body="{ data }">
-          <EditConfig :tenant="data" />
+          <div class="container">
+            <EditConfig :tenant="data" />
+            <DeleteTenant :tenant="data" />
+          </div>
         </template>
       </Column>
       <Column
@@ -96,6 +99,7 @@ import { storeToRefs } from 'pinia';
 import { TABLE_OPT, API_PATH } from '@/helpers/constants';
 import { formatDateLong } from '@/helpers';
 import EditConfig from './editConfig/editConfig.vue';
+import DeleteTenant from './deleteTenant/DeleteTenant.vue';
 import MainCardContent from '@/components/layout/mainCard/MainCardContent.vue';
 import RowExpandData from '@/components/common/RowExpandData.vue';
 
@@ -148,3 +152,8 @@ const filter = ref({
 // necessary for expanding rows, we don't do anything with this
 const expandedRows = ref([]);
 </script>
+<style scoped>
+.container {
+  display: flex;
+}
+</style>
--- a/services/tenant-ui/frontend/src/components/innkeeper/tenants/deleteTenant/DeleteTenant.vue
+++ b/services/tenant-ui/frontend/src/components/innkeeper/tenants/deleteTenant/DeleteTenant.vue
@@ -1,0 +1,53 @@
+<template>
+  <div>
+    <Button
+      :title="$t('tenants.settings.editSettings')"
+      icon="pi pi-trash"
+      class="p-button-rounded p-button-icon-only p-button-text"
+      @click="openModal"
+    />
+    <Dialog
+      v-model:visible="displayModal"
+      :style="{ minWidth: '500px' }"
+      :header="'Delete Tenant'"
+      :modal="true"
+      @update:visible="handleClose"
+    >
+      <ConfirmTenantDeletion
+        :tenant="props.tenant"
+        @success="$emit('success')"
+        @closed="handleClose"
+      />
+    </Dialog>
+  </div>
+</template>
+
+<script setup lang="ts">
+// Types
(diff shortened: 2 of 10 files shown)


services/tenant-ui/frontend/src/components/innkeeper/tenants/deleteTenant/DeleteTenant.vue
```diff
@@ -0,0 +1,53 @@
...
+        @success="$emit('success')"
+        @closed="handleClose"
+      />
+    </Dialog>
+  </div>
+</template>
+
+<script setup lang="ts">
+// Types
+import { TenantRecord } from '@/types/acapyApi/acapyInterface';
```
jamshale, line 27: I think we should simplify the import categories here. Like you did in the last PR.

services/tenant-ui/frontend/src/components/innkeeper/tenants/deleteTenant/DeleteTenant.vue
```diff
@@ -0,0 +1,53 @@
...
+defineEmits(['success']);
+
+// Props
+const props = defineProps<{
+  tenant: TenantRecord;
+}>();
+
+const displayModal = ref(false);
+const openModal = async () => {
+  // Kick of the loading asyncs (if needed)
```
jamshale, line 46: IMHO these comments aren't useful. I know there is other places with the same comments.

jamshale:
The code looks good. 👍 

I'm thinking for the soft delete, instead of having a boolean type flag, it should be a timestamp. You can still tell the record was soft deleted, if it has a timestamp or not. But then you know when the record was soft deleted which can be used for the hard delete logic. Say we want to keep soft deleted records for one month.

gurjmatharu:
> The code looks good. 👍
> 
> I'm thinking for the soft delete, instead of having a boolean type flag, it should be a timestamp. You can still tell the record was soft deleted, if it has a timestamp or not. But then you know when the record was soft deleted which can be used for the hard delete logic. Say we want to keep soft deleted records for one month.

Would a last updated work? It could be a universal last update column for any changes.

jamshale:
@gurjmatharu I guess that could work. I was thinking just a new column for `deleted_at` or something like that. Then you can just filter by if `delete_at == None` for the active tenants. As long as there's something saved for the time of deletion it should be fine though.

loneil:
> The code looks good. 👍
> 
> I'm thinking for the soft delete, instead of having a boolean type flag, it should be a timestamp. You can still tell the record was soft deleted, if it has a timestamp or not. But then you know when the record was soft deleted which can be used for the hard delete logic. Say we want to keep soft deleted records for one month.

I would keep the state deleted. All the other records in aca-py to my knowledge keep the state to track active/deleted/others and then filter on that so I think it's good to keep the same convention for this one.

loneil:
> @loneil So what would you do for the hard delete? Just delete everything that has been soft deleted?
> 
> You could still have the same state convention with the soft deleted time. But if it's not important than it can be kept this way.

Yeah mostly just wanting to keep the state for filtering and choices since it's a convention. For hard deletes, not sure there's really any concrete plans for how that would be handled but yeah knowing when it was deleted would be useful of course for whatever the plan ends up being (if there is one eventually). The existing Updated At probably wouldn't be used again but I'm not sure if there's something that could update that on a deleted tenant and make it appear to be deleted more recently than it was.

services/tenant-ui/frontend/src/components/innkeeper/tenants/Tenants.vue
```diff
@@ -33,7 +33,10 @@
       <Column :expander="true" header-style="width: 3rem" />
       <Column :sortable="false" :header="$t('common.actions')">
         <template #body="{ data }">
-          <EditConfig :tenant="data" />
+          <div class="container">
+            <EditConfig :tenant="data" />
+            <DeleteTenant :tenant="data" />
```
loneil, line 38: @gurjmatharu can we block the delete if it's the Innkeeper Tenant? I think the only way to reliably know is to check `tenant_name` is `traction_innkeeper` since the IDs vary in each environment. Should be ok since that's the convention for Traction anyways.

While you're at it we can also disable the EditConfig gear there too since there's no need for Innkeeper issuance stuff. Been meaning to do that for a while.

image

gurjmatharu:
@loneil sure, I'll block the delete/edit.  Not sure how I'd disable them on the front end but the backend we can add a check.  So if the user does click the buttons there will be an error

@jamshale I'm going to also add a `deleted_at` that only gets written on `soft_delete`.

loneil:
> @loneil sure, I'll block the delete/edit. Not sure how I'd disable them on the front end but the backend we can add a check. So if the user does click the buttons there will be an error
> 
> @jamshale I'm going to also add a `deleted_at` that only gets written on `soft_delete`.

I mean just disable them on the frontend, probably don't need to do on the backend. Can just check the tenant_name from the record and then disable the button.

jamshale (approved):

PR #908, opened 2023-11-08
Update msw library and http mocks

jamshale:
Updated the msw library and the tests. Not sure if mocking the api this extensively was worth it lol. Might come in handy one day.

I removed the local storage tests. Vitest is using it differently now. Not sure why or how to properly test it.
I removed one test who's failed request wasn't working. Weird because the same pattern is working everywhere else. Not important enough to spend time on.


loneil (approved):
Awesome, thanks for taking a look. Looks good to me

PR #909, opened 2023-11-09
Remove null pointer in oca plugin schema

jamshale:
Not sure when this broke but the variable this schema reference must have been removed.

Weird it didn't throw errors.

Should consider running the `ruff` linting on the plugins to catch something like this.


esune (approved):

esune:
@jamshale is linting set-up in the plugins repo such that we could follow that example to lint plugins here that will not be moved?

jamshale:
@esune There is now yes. You install a poetry plugin and then navigate to the repo and then run `poetry run ruff check .`
It's pretty simple. My IDE picked this up right away though, because I have the vscode ruff extension installed.

jamshale:
@esune Actually the plugins in traction aren't using ruff yet. Maybe they should be updated to use ruff and then conform to the same rules as acapy like the plugin repo. I think this could wait until we remove the plugins that are already in the plugin repo.

PR #912, opened 2023-11-09
Create option to show soft deleted tenants

jamshale:
Added a toggle button to view the soft deleted tenants.

[Screencast from 2023-11-09 03:10:00 PM.webm](https://github.com/bcgov/traction/assets/31809382/0b39efb2-f8e9-4bfd-84b4-92eaebea7808)


loneil:
Nice, just popped into my mind it would be good to have a `restore` now that we have this (and it's just soft fields in the Tenant record). Would need another endpoint I think... (unless there's already a PUT on the tenant record that would do it) but can add a ticket for that later.
Think we could just cram a small button next to that DELETED badge in that case on this UI.

loneil (approved):
Looks great. Works on this PR

PR #914, opened 2023-11-10
Expose server settings to Innkeeper UI and Tenant UI via plugin routes

loneil:
### Plugin
Add in endpoints to the Innkeeper plugin to allow Innkeepers and Tenants to access ACA-Py server config.
There is a unsecured (for multitenancy, secured with x-api-key) endpoint `/server/status/config` so extend that in the Innkeeper and Tenant endpoints so it can be fetched with tenant JWTs. Also extend it/remove things as appropriate.

Keep the "/server/status/config" pathing, even if it's a bit verbose in these 2 new spots

`/innkeeper/server/status/config`
image

Add aca-py version, ensure innkeeper wallet key does not get returned.

`/tenant/server/status/config`
image

Add version, remove a number of things a tenant does not need to access.
See `plugins/traction_innkeeper/traction_innkeeper/v1_0/tenant/routes.py` in files changed for details.

### Tenant UI - Innkeeper

Add a "Server Configuration" page for the Innkeeper that displays a dashboard of 
- These new exposed ACA-Py settings
- Tenant UI settings
- Plugin list (unsecured existing aca-py endpoint)
- Other misc  

Break it down into some visual sections for important things. 
Allow full settings view with expanders at the bottom
image

### Tenant UI - Tenant
Just show the raw data on the Settings page.
I'm not sure how useful this is to a Tenant, but they might need to troubleshoot the default settings vs their own settings, so put it somewhere they can see it.

Maybe we could show "My Setting" vs "Default Setting" for each of these settings on the UI or something in the future... can decide UX needs as we go, but the data is there now.

image

### Other
Clean up some hard-coded values we don't need from config
- Ledgers are now all listed out to the Tenant/Innkeeper in other spots
- version is now fetched)


plugins/traction_innkeeper/traction_innkeeper/v1_0/innkeeper/routes.py
```diff
@@ -917,6 +919,41 @@ async def innkeeper_authentications_api_delete(request: web.BaseRequest):
...
+@error_handler
+async def innkeeper_config_handler(request: web.BaseRequest):
+    context: AdminRequestContext = request["context"]
+    # use base/root profile for server config, use Tenant Manager profile
+    # this is to not get the Innkeeper tenant's config, but the server cfg
+    mgr = context.inject(TenantManager)
+    profile = mgr.profile
+
+    config = {
+        k: (
```
jamshale, line 934: This is nice, but I like things to be more readable than shorter. I'm assuming that `k` = `key`. Imho is better just to write key here so other people don't need to decipher it.
It's not a big deal or anything but saving a couple characters doesn't seem worth it.

services/tenant-ui/frontend/src/components/innkeeper/config/ServerConfig.vue
```diff
@@ -0,0 +1,242 @@
...
+        </div>
+
+        <div class="col">
+          <!-- Details about tenants/reservations -->
+          <Fieldset
+            :legend="$t('serverConfig.tenants.title')"
+            :toggleable="true"
+          >
+            <p class="mt-0">
+              <strong>{{ $t('serverConfig.tenants.token') }}</strong>
```
jamshale, line 91: Could probably have a reusable component for these fields. Would make this component cleaner.

jamshale (approved):
Looks good. Couple small comments but nothing important.

services/tenant-ui/frontend/src/components/innkeeper/config/ServerConfig.vue
loneil: You thinking a reusable component for the `p>strong>field name: fieldvalue` pairs with this?
Or breaking the sub panel contents into their own components to keep the ServerConfig page smaller?

(both sound like reasonable ideas)

plugins/traction_innkeeper/traction_innkeeper/v1_0/innkeeper/routes.py
loneil: Good idea

services/tenant-ui/frontend/src/components/innkeeper/config/ServerConfig.vue
jamshale: I was just thinking the `<p>strong>field name: fieldvalue` when I wrote this. Just because I think it would look cleaner having so many strong and curly brace elements. You could make the value prop undefinable as well and avoid all the `?'s` in this component.

loneil: Extracted to a `ConfigItem` component that can take in text (or other simple types) as props, or slots if HTML (icons, links etc) want to be provided. Added the `:` delimiter to i18n so it can be changed in one spot if we ever needed to change up the format. Can add this to a bunch of other places around the app as we go.
https://github.com/bcgov/traction/pull/914/commits/dcb4c2c477c3b3bf5444cf459bff417633a83223

jamshale: Nice! That looks better.

jamshale (approved):

PR #939, opened 2023-12-01
Add support toggling json view of schemas in CreateSchemaForm

### LINKED ISSUE #921: In the tenant UI allow schema creation with a JSON blob
As a Tenant UI Issuer, I want to be able to toggle over to a JSON entry mode on the schema modal, so that I can easily enter a schema from documentation or training.
--

AC:
- Like on the issuance, there's a toggle for JSON entry on the schema popup
- When you toggle back and forth from JSON to fielded entry, it keeps the 2 in sync
- Invalid JSON gives a warning message, same as on issuance.

Gavinok:
This pr resolves #921 by introducing a standard way to toggle between the json representation of a schema as seen below.

- [X] Like on the issuance, there's a toggle for JSON entry on the schema popup
- [X] When you toggle back and forth from JSON to fielded entry, it keeps the 2 in sync
- [X] Invalid JSON gives a warning message, same as on issuance*.

*: For consistency invalid JSON is not allowed in this new component. How this error is indicated will be handled by the parent components.

Regular representation:

JSON representation:

The two are always kept in sync when switching. This way users can enter
some fields via the JSON view and other fields using the regular list view.

Here is an example of the attribute list being updated and reflected in the regular list view.

Change made in the JSON view:

Resulting change seen in the list view:

To reduce duplication of code I have adapted the code used in `EnterCredentialValues.vue` and extracted the functionality of toggling to a JSON view into it's own common component `ToggleJson`.


loneil:
@Gavinok could request a review from @jamshale if needed while I'm away.

jamshale:
This is a really good feature. We should have another ticket to do this for credential definitions as well.

services/tenant-ui/frontend/src/components/common/ToggleJson.vue
```diff
@@ -0,0 +1,60 @@
...
+    if (res) {
+      valuesJson.value = res;
+    } else {
+      showRawJson.value = !showRawJson.value;
+    }
+  } else {
+    if (!props.fromJson(valuesJson.value)) {
+      showRawJson.value = !showRawJson.value;
+    }
+  }
```
jamshale, line 58: If you're like me and hate unnecessary curly braces you can remove a lot of these when there's only a single statement.

jamshale (dismissed):
Really nice work!

Unit tests are annoying, but some could be considered. Especially in the ToggleJson component which will likely be reused or changed in the future.

Gavinok:
> Really nice work!
> 
> Unit tests are annoying, but some could be considered. Especially in the ToggleJson component which will likely be reused or changed in the future.

Sure thing I'll try adding them now

jamshale (approved):

PR #959, opened 2023-12-19
Add email contact to tenants

### LINKED ISSUE #900: Update email address for approved tenant
Add functionality to update the recorded contact email for approved/registered tenants: this is useful if the contact person responsible for the tenant changes, and communications need to be routed to someone else.

Gavinok:
This PR resolves #900

Emails are now tied to a tenant upon creation and can be viewed in the Profile page

In the Settings tenants can change the email associated with their tenant

Finally the innkeeper can view this email when inspecting the raw content of a tenant


services/tenant-ui/frontend/src/plugins/i18n/locales/en.json
```diff
@@ -326,6 +326,7 @@
     "imageUrl": "Image URL",
     "issuer": "Issuer",
     "name": "Name",
+    "contact": "Email",
```
loneil, line 329: Think I'd make this "Contact Email" explicitly so if one of multiple users see the profile it's a little clearer it's the contact email for the Tenant and not their personal email.

loneil, line 329: Think I'd make this "Contact Email" explicitly so if one of multiple users see the profile it's a little clearer it's the contact email for the Tenant and not their personal email.

plugins/traction_innkeeper/traction_innkeeper/v1_0/innkeeper/tenant_manager.py
```diff
@@ -168,6 +172,8 @@ async def create_tenant(
                     auto_issuer=auto_issuer,
                 )
                 await tenant.save(session, reason="New tenant")
+                new_tenants = await tenant.query(session)
+                self._logger.error(f"The current list of tenants is {new_tenants}")
```
loneil, line 176: Not sure if this is meant to be left in, or at level `error`?

loneil (commented):
Looks good, I like the flow that it takes the contact email from the reservation and then the tenant record itself (rather than the reservation) is what gets updated by the user themself.

Couple minor comments in code.

One thing we could also consider I think is adding the contact_email to the table itself so the Innkeeper can see it at-a-glance? Unless anyone else reading doesn't think that's useful for the screen real estate.

plugins/traction_innkeeper/traction_innkeeper/v1_0/innkeeper/tenant_manager.py
Gavinok: Meant to drop this logging. Thanks for catching that

Gavinok:
The innkeeper can now view and contact tenants email's via the tenant view



The link is a `mailto:` href so innkeepers will have their default mail client open when the link is clicked

loneil:
> The innkeeper can now view and contact tenants email's via the tenant view
> 
> 
> 
> The link is a `mailto:` href so innkeepers will have their default mail client open when the link is clicked

@Gavinok oh nice call!

Minor-est quibble but can we move the email column one over to the left, just because (unless there's additional row actions like the schema one) all the created/updated dates are the rightmost column in the tables.

Gavinok:
@loneil how does this look?

loneil (approved):
Looks good

Issue #935, opened 2023-11-29
Scope Tenant UI tests to be executed only when code changes


esune:
The tenant UI tests are currently always executed on new PRs, buut they should be scoped to run only when changes happen in the `service/tenant-ui` folder.

See https://github.com/bcgov/traction/blob/main/.github/workflows/test_tenant_ui.yml for job definition.

PR #963, opened 2023-12-20
Adding scope to the job to run only when changes happen in the service/tenant-ui folder.

rajpalc7:
Resolves https://github.com/bcgov/traction/issues/935


esune (approved):
👍🏻

PR #971, opened 2024-01-02
Restore soft-deleted Tenant as Innkeeper

loneil:
Endpoint to restore a soft-deleted Tenant for Innkeepers
image

UI in the Innkeeper table
image


services/tenant-ui/frontend/src/components/innkeeper/tenants/deleteTenant/ConfirmTenantDeletion.vue
```diff
@@ -54,19 +56,19 @@ const toast = useToast();
...
-  emit('closed');
+  try {
+    await innkeeperTenantsStore.deleteTenant(props.tenant.tenant_id);
+    toast.success(
+      t('tenants.settings.confirmDeletionSuccess', [props.tenant.tenant_name])
+    );
+    emit('success');
+    emit('closed');
+  } catch (err: any) {
+    console.log(err);
```
jamshale, line 70: Could probably use console.error here

jamshale (approved):

services/tenant-ui/frontend/src/components/innkeeper/tenants/deleteTenant/ConfirmTenantDeletion.vue
loneil: Ah yeah, looks like there's a few around that are log in catch block that could be better as error. I'll bulk update them in a next PR.

PR #974, opened 2024-01-04
[Feature] Enable SMTP Authentication in Tenant UI

i5okie:
## Enable using authentication with SMTP

Nodemailer used in Tenant UI supports using authentication. This PR extends the code to allow setting `secure`, and providing `user` and `password` variables via the Helm values. 

_For additional information refer to the `nodemailer` documentation: https://nodemailer.com/smtp/#authentication_

Thanks @loneil for your help!


loneil:
To test this on the deployed PR we can temporarily set `reservation.auto_approve` in from `plugin-config.yml` in the config-map to false (or update it in values-pr to false and redeploy). Then reservation requests on the PR will send emails.

Can test out when deployment goes through but looking at code it seems ok. Probably good global setting to have though don't think we'll ever need this as we'd likely always be using the gov smtp server in OCP where it doesn't need auth?

esune (approved):
Looks good assuming the testing with no auth credentials yields positive results. @loneil it's true that for BCGov OCP we would not need the auth credentials, but it's nice to have the option 😉

i5okie:
Works. Tested with `reservation.auto_approve` set to false as per @loneil's comment.

loneil (approved):
Looks good! Tried out locally on VPN with secure false to confirm emails send

PR #1016, opened 2024-02-12
Release 0.4.0

esune:
Release version 0.4.0.

Once approved, a release tag will be created before merging back to main.


loneil:
I updated the Tenant UI version labels here. If we didn't want to do that let me know and I'll rebase revert it out.

loneil (approved):

esune:
> I updated the Tenant UI version labels here. If we didn't want to do that let me know and I'll rebase revert it out.

I did search/replace and missed that - thanks for catching it.
I synced with the latest changes in `main` which include the DRPC plugin.

PR #1050, opened 2024-03-27
Remove bcovrin-dev, set second (aliased) bcovrin-test for local multiledger

loneil:
As bcovrin-dev is deprecated we can't use it any more in Traction. So handle that in a couple ways

**Deployments**'
For dev and PR deployments just remove BCovrin-dev as an option. These have other combos of candy and sovrin so they have other ledgers as well as bcovrin-test.

**Default helm values**
Just remove bcovrin-dev, leave only bcovrin-test.
Default helm is overridden everywhere we use it and there's the dev and pr examples in here of how to specify multiple ledgers if someone needs them.

**Local Docker**
We do want multiple ledgers on local dev for multi-ledger development and tenant UI development.
Not going to use candy or sovrin on local dev. We stand up 2 local endorser instances in the docker environment, so just have each of those using the same ledger (bcovrin-test), but alias the IDs and endorser details differently so it's interpreted as "multiple" ledgers.
Replace bcovrin-dev with "bcovrin-test-1" (all endorser setup is "endorser-1" etc etc, so keep that convention).

Tried this out locally and it's working as expected.


loneil:
Tagged a bunch of people on this in case anyone has any thoughts/opinions/gotchas they can think of.

esune (approved):

Issue #1049, opened 2024-03-26
Update Helm Chart to add ArgoCD deployment compatibility


i5okie:
The helm chart must be updated to be deployed with ArgoCD.
ArgoCD uses `helm template` command when deploying, rendering logic, like secret lookup inoperable. If deployed as is with ArgoCD, any existing secret will be overwritten.

PR #1055, opened 2024-04-04
[Helm] Allow specifying existingSecret for all secrets

i5okie:
### Updating the Helm chart to allow specifying an `existingSecret` for any secret used for Traction.

Updated Values:
```yaml
  secret:
    adminApiKey:
      existingSecret: ""
    walletKey:
      existingSecret: ""
    pluginInnkeeper:
      existingSecret: ""
```

Addresses: https://github.com/bcgov/traction/issues/1049


charts/traction/templates/_helpers.tpl
```diff
@@ -87,6 +88,33 @@ Return true if a database secret should be created
...
+Return true if a api secret should be created
+*/}}
+{{- define "acapy.api.createSecret" -}}
+{{- if not .Values.acapy.secret.adminApiKey.existingSecret -}}
+{{- true -}}
+{{- end -}}
+{{- end -}}
+
+{{/*
+Return true if a api secret should be created
```
esune, line 101: Nit.: this is creating the wallet secret

esune (changes requested):
Small tweak in description required, but looks good otherwise.

charts/traction/templates/_helpers.tpl
i5okie: has been addressed in the following commit

esune (approved):

PR #1059, opened 2024-04-08
feat: RPC plugin demo

amanji:
This PR adds a new demo, showcasing the RPC plugin and also serves as a general LOB example. Since it contains many of the same features as the existing plugin demo, such as automatic tenant reservation and approval, automatic webhook registration, etc., it effectively supersedes the existing demo (which has therefore been removed as a part of this PR) 

In the future this demo folder can and should be expanded to include other examples.

Documentation has been included.


loneil:
The build/deploy checks are all failing as this is a PR from a fork so it won't run any of the steps on the BC Gov OCP infra. Which is fine for this change as it's specifically to the demo folders "outside" the Traction app so none of it is built or deployed for PRs anyways.

So can be undrafted/marked Ready for review if it's ready without worrying about the branch/build.

PR #1060, opened 2024-04-08
Correct ToggleJson support for creating schemas

### LINKED ISSUE #1030: Tenant UI: JSON toggle on Create Schema is broken
No matter what I input, toggling the JSON switch throws an invalid error, nothing in console log.

image

Having similar issue in the Offer Credential feature (complaining about an invalid json). Should fix that with this.

Gavinok:
This PR resolves #1030

Uncertain as to why this is necessary but for whatever reason showRawJson defaulting to a false value seemed to never be properly applied and needed to be defaulted to true. For this reason it's usage has been inverted and the variable has been renamed to displayAsForm.

### DIFF
--- a/services/tenant-ui/frontend/src/components/issuance/schemas/CreateSchemaForm.vue
+++ b/services/tenant-ui/frontend/src/components/issuance/schemas/CreateSchemaForm.vue
@@ -74,7 +74,6 @@ import ToggleJson from '@/components/common/ToggleJson.vue';
 const toast = useToast();
 const { t } = useI18n();
 
-const other = ref(false);
 const governanceStore = useGovernanceStore();
 const { loading, selectedSchema } = storeToRefs(useGovernanceStore());
 
--- a/services/tenant-ui/frontend/src/components/issuance/schemas/CreateSchemaForm.vue
+++ b/services/tenant-ui/frontend/src/components/issuance/schemas/CreateSchemaForm.vue
@@ -204,7 +204,7 @@ function jsonToSchema(jsonString: string): SchemaSendRequest | undefined {
     formFields.version = parsed.schema_version;
     return parsed;
   } else {
-    toast.error('The JSON you inputted has invalid syntax');
+    toast.error('Invalid JSON detected');
     return undefined;
   }
 }
--- a/services/tenant-ui/frontend/src/components/issuance/schemas/CreateSchemaForm.vue
+++ b/services/tenant-ui/frontend/src/components/issuance/schemas/CreateSchemaForm.vue
@@ -54,7 +54,6 @@
 
 <script setup lang="ts">
 // Libraries
-import Tester from '@/components/common/./Tester.vue';
 import { useVuelidate } from '@vuelidate/core';
 import { helpers, required } from '@vuelidate/validators';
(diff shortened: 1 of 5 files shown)


loneil (changes requested):
When opening the schema screen it looks like it's toggled over to JSON to start when it's on the form mode by default

image

Gavinok:
For some reason the input switch seems to consider it's self to always be invalid. As far as the direction of the button this seems to only update displayAsForm when it is defaulted to true. I will poke around at the internals of primevue to see if I can get an explanation to why.

Gavinok:
I have found the culprit 

https://github.com/bcgov/traction/blob/restore-json-toggle-schema/services/tenant-ui/frontend/src/assets/primevueOverrides/inputSwitch.scss#L3-L22

@loneil From what I can tell this is taking advantage of the `.p-inputswitch-checked` class which isn't being used by the input switch at all. 

If you look at the profile settings the same issue is occurring



I modified the code to now use `[data-p-highlight="true"]` as the checked indicator. I also made some smaller changes which should make the code a lot closer to the original.

loneil:
Ok I'd guess that `p-inputswitch-checked` class got removed at some point in a Primevue version change.

services/tenant-ui/frontend/src/components/issuance/schemas/CreateSchemaForm.vue
```diff
@@ -54,6 +54,7 @@
 
 <script setup lang="ts">
 // Libraries
+import Tester from '@/components/common/./Tester.vue';
```
loneil, line 57: Is this a leftover?

services/tenant-ui/frontend/src/components/issuance/schemas/CreateSchemaForm.vue
```diff
@@ -204,17 +206,19 @@ function jsonToSchema(jsonString: string): SchemaSendRequest | undefined {
...
 // Form submission
 const submitted = ref(false);
 const handleSubmit = async (isFormValid: boolean) => {
   submitted.value = true;
-
+  console.log('trying to validate');
   try {
+    console.log('is valid?');
     if (!isFormValid) return;
+    console.log('valid');
```
loneil, line 221: Probably don't need these logs

loneil (commented):
Working for me on this PR, just a couple comments

services/tenant-ui/frontend/src/components/issuance/schemas/CreateSchemaForm.vue
Gavinok: Both are just leftovers from testing the colour toggling. I have now removed them. Thanks for the review!

services/tenant-ui/frontend/src/components/issuance/schemas/CreateSchemaForm.vue
```diff
@@ -74,6 +74,7 @@ import ToggleJson from '@/components/common/ToggleJson.vue';
 const toast = useToast();
 const { t } = useI18n();
 
+const other = ref(false);
```
loneil, line 77: Another one

loneil (approved):

PR #1063, opened 2024-04-11
feat: add tenant id to log output

amanji:
This PR modifies the Tenant plugin with a tenant-specific log formatter that will inject the Tenant ID into log records (replacing the default Wallet ID in the ACA-Py multitenant setup).

Tenant ID is dependent on the Wallet ID which itself is dependent on the auth token present in request headers. As a result, only tenant-authorized methods and routes will log the Tenant ID.


amanji:
Will rebase this onto a bcgov branch

PR #1064, opened 2024-04-11
feat: RPC plugin demo

amanji:
This PR adds a new demo, showcasing the RPC plugin and also serves as a general LOB example. Since it contains many of the same features as the existing plugin demo, such as automatic tenant reservation and approval, automatic webhook registration, etc., it effectively supersedes the existing demo (which has therefore been removed as a part of this PR)

In the future this demo folder can and should be expanded to include other examples.

Documentation has been included.


demo/plugins/rpc/app/index.html
```diff
@@ -0,0 +1,13 @@
+<!DOCTYPE html>
+<html lang="en">
+  <head>
+    <meta charset="UTF-8">
+    <link rel="icon" href="/favicon.ico">
+    <meta name="viewport" content="width=device-width, initial-scale=1.0">
+    <title>Vite App</title>
```
jamshale, line 7: Not sure if this gets changed but if not than a better name would be nice.

jamshale (dismissed):
Mostly just scanned over but looked really good.

loneil (dismissed):
Reorg looks good to me (and don't need to rebase onto a bcgov branch as OCP building not needed for this)

PR #1074, opened 2024-04-18
chore: moves tenant lob demo out of core services

amanji:
Tenant LOB is not needed to run core services and has been moved to the demo folder


loneil (approved):
Great!

PR #1118, opened 2024-04-23
Non-Traction service library updates

loneil:
Non-Traction lib updates


swcurran:
Did you do this manually, or is there a tool you can use to apply all of the updates?  The “one at a time” dependabot approach is necessary but painful. It would be really helpful if there was a way to have a single dependabot PR that applies all the updates that come along.  Or better — a script you could run on a local package file to get updates for all of the depdendencies. Testing them all at once would be good.

loneil:
Manual. For dependabot we really need something that packs them into 1 PR that can be reviewed and looked over as needed, as I don't find having 10-20 PRs get spammed particularly helpful. (although if someone isn't keeping up with things as a task it warns at least, but with the current setup I think we're getting into noise territory over necessary security warnings, planning on brining this up at a retro) But I don't know if anything dependabot supports that...

For `npm` style updates the thing I typically do is a regular
`npx npm-check -u` to see updates available and test them out. This is regardless of any dependabot setup but that will resolve them as a group when I do my own PR after that.
A lot of times this will require changes to FE, unit tests, etc (so IMO we can't just rely on any automerging of dependabot, though unit tests catch a lot) to accommodate changes in libraries.

I'm not sure if there's anything equivalent to `npm-check` for PyPI

PR #1119, opened 2024-04-24
Update Traction to ACA-Py 0.12.0

loneil:
Update the ACA-Py docker image to 0.12.0

Update the Innkeeper plugin poetry dependencies to 0.12.0 (and add anoncreds dependency)

**Note** This does not pull updated versions of the _other_ plugins from the plugins repo at 0.12.0. Those 3 (multitenant provider, basic message, connection update) are `main`, referencing 0.11.0 aca-py PyPI library at this time.
Options for that are
- Just do that for now (those plugins still referencing 0.11.0 library)
- Wait for tagging solution on the plugins repo
- Fork and just handle those 3 plugins


jamshale (approved):
Think that this is fine for now. Not installing the `aca-py` extras from the plugin repo and relying on the base image and the deps from innkeeper plugin. Will be able to use tags and upgraded plugins soon.

Issue #1107, opened 2024-04-19
Troubleshooting on demo rpc app startup


loneil:
Adding notes on some additional troubleshooting docs that can possibly be added from trying out steps here
https://github.com/bcgov/traction/tree/main/demo/plugins/rpc

**Issue with .venv**
On startup of the devcontainer on my machine it didn't seem to be creating venv.
So documented step of `source .venv/bin/activate` was erroring
![image](https://github.com/bcgov/traction/assets/17445138/f771f1f8-9c5c-4850-822a-0476577514dc)

had to create a .venv 
![image](https://github.com/bcgov/traction/assets/17445138/711a13cb-2124-4782-b3ba-1fdc25f7adfb)
then it worked

also to start up debugging had to set a python intepreter in vscode
- command pallette (ctrl shift v) > Select Interpreter
  - Enter interpreter path
  - find
  - select .venv that was created

**Copy Sample env to .env**
- Documentation note to create an .env file
- Check default Traction ports in the sample file

amanji:
Resolved by #1155

PR #1155, opened 2024-05-13
chore: update demo docs for clearer setup

amanji:


amanji:
Resolves: #1107

demo/plugins/rpc/README.md
```diff
@@ -14,10 +14,12 @@ To initialize the Controller, run the following commands from the root of this d
 
 ```sh
 cd controller/
-source .venv/bin/activate
 poetry install --no-root --all-extras
+cp .env.example .env
 ```
 
+If you are running Traction in a devcontainer you should adjust the `PROXY_URL` in the `.env` to point to `host.docker.internal` instead of `localhost`.
```
loneil, line 21: For me (on Windows) I had to change to host.docker.internal even though I wasn't running Traction itself in a devcontainer (it's funny, was just sorting through the same thing on an unrelated issue https://github.com/hyperledger/aries-cloudagent-python/issues/2938)

I _think_ that's Windows-docker-specific and not just the setup I'm using so maybe worth a language in this part about "If you are running Traction in a devcontainer **(or on Windows outside WSL)**" or something?

loneil (approved):
Looks good! Followed through and it worked.
Minor note about Windows-specific startup

demo/plugins/rpc/README.md
amanji: Updated for clarity

loneil (approved):

Issue #1182, opened 2024-05-30
Spaces in WebHook URL cause controller calls to fail


jleach:
While I get that this is operator error, it would be helpful if when entering a WebHook URL the Web UI stripped any starting or trailing whitespace that may come from a copy-and-paste error. They cause controller calls to fail and also are hard to catch and debug.

loneil:
https://github.com/bcgov/traction/pull/1199 merged

PR #1197, opened 2024-06-04
Trim whitespace from webhook urls. Dependency updates

loneil:
Address https://github.com/bcgov/traction/issues/1182

Also do dependency updates for everything.

Add temporary workaround DCO check until (if?) the hosted one we're all using gets fixed
https://github.com/dcoapp/app/issues/211


loneil:
Closing this and reopening to get DCO back

PR #1256, opened 2024-07-12
Use BasicMultitokenMultitenantManager for multitenant_provider

jamshale:
Using the default `MultitenantManager` with the `multitenant_provider` plugin will create separate DB's per tenant.


jamshale:
Changing this in favour of a full PR.

PR #1280, opened 2024-07-26
Tenant UI 2.0 protocol changes

loneil:
Update Tenant UI credential/proof/etc protocols that can are called through the interface to use 2.0 protocols.
Update unit tests, some other Typescript refactors. Dependency updates.

Tested connecting OOBs, issuing Traction Tenant to Traction Tenant, accepting/denying revoking credentials as issuer and holder, sending pres reqs etc. Tested out with BC Wallet as well.


esune (approved):
👍🏻 

Only wondering if we should add something to the backlog and/or TODOs in the code to handle presentation exchanges that are not of type `indy` (i.e.: `dif`) in the future.

loneil:
> 👍🏻
> 
> Only wondering if we should add something to the backlog and/or TODOs in the code to handle presentation exchanges that are not of type `indy` (i.e.: `dif`) in the future.

Right, probably some other UI considerations around using non Indy through the interface. Can track here https://github.com/bcgov/traction/issues/1284

PR #1290, opened 2024-07-30
Update action to delete old images

esune:
Updated `snok/container-retention-policy` action to latest version, updated workflow to match new action API.

I am not 100% sure that `tag-selection: untagged` will pick up the images that don;t have a human-readable tag - it looks like the images named after the commit sha are also tagged as such. If that is the case we'll have to think about how to clean-up more images and leave the explicitly tagged ones (releases, PRs).


loneil (approved):
👍

PR #1292, opened 2024-07-30
Delete closed PR images

esune:
Update action to correctly delete images when PR is closed


loneil (approved):
I think this will work? Not too familiar with the action step

PR #1294, opened 2024-07-31
Support fully delete/drop tenant and associated wallet

### LINKED ISSUE #1253: Innkeeper: fully delete/drop tenant and associated wallet
An innkeeper should be able to manage tenants and drop/delete a tenant and its wallet if necessary.
Each tenant should be able to also do so independently.

Gavinok:
This PR resolves #1253

I have added acapy endpoints for hard deleting tenants for both the innkeeper to delete any tenant and the tenant to delete themselves. I have also exposed this function in the UI

For the innkeeper it's simply a check box

For the tenant a new button has been added to the profile page which will then prompt the user for confirmation.


loneil:
We're having the "you must type your name" super-confirm on just soft deleting a tenant as the Innkeeper, so I think we should for sure have more than just the standard confirm Yes/No on the _Tenant_ Profile page.

There was discussion in the past about having the soft-delete flag be done first and then after a certain amount of time enabling a permanent delete. But if we're exposing the endpoint to permanent delete then of course the tenant can call it themselves. (I guess that endpoint could check the soft-delete flag was set at the least)

loneil:
Chatted with @esune we can do this:

1. On the Innkeeper side minor tweak to the UI to have radio buttons as "Suspend Tenant" and "Permanently Delete". If they click permanent delete display a warning message with an icon in bold yellow saying what it will do above the button.
2. On the Tenant Profile page, use the same Confirmation box with those 2 radio buttons. So a Tenant has to manually type in their tenant name and has to pick which deletion type
3. Allow the Tenant in the above box to soft-delete (the Suspend option). Think this should be possible for the tenant, it's just setting a flag in their tenant record. Will likely need to add a version of the DELETE `/innkeeper/tenants/{id}` endpoint that the Tenant can call, or first can check if setting the delete flag can be done with PUT `/tenant/wallet` can be used for this?

The soft delete is really just setting a flag at this time so we are adding a couple tickets for enhancement to that behaviour that can be picked up after this is done (can proceed with the above changes just using the existing soft delete behaviour)

Gavinok:
Updated the ui and added support for soft deleting tenants

Here is the warning shown when permanently delete is selected



This same prompt is now also used for tenants as well

esune (commented):
Tested the UI and it looks good to me. Confirmed a suspended tenant can still sign-in and "do stuff", for which we have follow-up tickets scheduled.

Only wondering if we should change the color of the delete button to be red to highlight the dangerous operation, as well as making it not as wide as the screen.

Gavinok:
I had thought about making it red as well but figured all the warnings would deter people from accidentally clicking it.  I will change it now

loneil:
Yes let's fix the width of this button for sure

image

Gavinok:
Updated tenant deletion button

loneil:
Not exactly related to this PR but just confirming data deletion after tenant hard delted

image

Note this is for the current Traction mode of all wallets in the single database, going to check locally running in multi-DB mode.

esune (approved):
Good to go on my end! 👍🏻

loneil (approved):

PR #1309, opened 2024-08-10
Node version update

loneil:
Tenant UI running on Node 18 for build and backend API, which is in maintenance (starting getting some deprecation warnings in libraries) so update to current release Node 22. https://nodejs.org/en/about/previous-releases


.github/workflows/test_tenant_ui.yml
```diff
@@ -18,7 +18,7 @@ jobs:
 
     strategy:
       matrix:
-        node-version: [18.x, 20.x]
```
esune, line 21: Do we need to keep Node 20 in the matrix if we are basing the code on Node 22?

loneil: Not really for us, just that 20 is still active until later in the fall so if anyone else were deploying or anything on 20 this shows a sanity check that it tests out. Doesn't make much difference though

esune: Gotcha. My thought is that whoever takes the code will be doing it on the "current" state and Node release, and removing the older version for which we would not provide support might also speed up the tests (not that they are slow currently). Wondering if it would be a better choice to use current + next release in the matrix to catch potential issues with the new versions early?

loneil: Yeah next one is better I think. No 23/24 support on that action yet it seems so siloing to 22.x right now, can add 23 once that odd-preview gets there this fall.

esune (approved):
👍🏻

PR #1330, opened 2024-08-19
Improvements to Tenant Deletion

esune:
- Improve UI for deleting suspended tenants
- More consistent wording for tenant deletion
- Block API usage for deleted tenants
- Delete API keys when deleting or suspending tenants 

Opening on behalf of @Gavinok


plugins/traction_innkeeper/traction_innkeeper/v1_0/innkeeper/routes.py
```diff
@@ -522,6 +532,60 @@ async def tenant_create_token(request: web.BaseRequest):
...
 
 
+@docs(
+    tags=["multitenancy"],
+    summary="Get auth token for a subwallet (innkeeper plugin override)",
+)
+@request_schema(CreateWalletTokenRequestSchema)
+@response_schema(CreateWalletTokenResponseSchema(), 200, description="")
+@error_handler
+async def tenant_wallet_create_token(request: web.BaseRequest):
```
loneil, line 542: So for discussion and any other people's thoughts on what we are doing here (correct me if I'm wrong on any of this)

- We want to stop token creation for "suspended" soft-deleted tenants (IE tenants in the deleted state)
- Token creation is overridden in the other plugin Multitenant Provider
- See that here https://github.com/hyperledger/aries-acapy-plugins/blob/main/multitenant_provider/multitenant_provider/v1_0/routes.py#L76
- We can't put the tenant state check there because in that plugin there isn't any "Tenant" concept, that's Innkeeper plugin and obviously the multitenant provider could be used independently (why use that plugin, see details here https://github.com/hyperledger/aries-acapy-plugins/tree/main/multitenant_provider)

I guess one concern of this here is we are duplicating some code essentially, if some enhancement or fix happens in the other plugin create_token we wouldn't get it here?

There's some other code in the plugin at this line:

https://github.com/hyperledger/aries-acapy-plugins/blob/main/multitenant_provider/multitenant_provider/v1_0/routes.py#L116

where `if (not wallet_record.requires_external_key) and wallet_key:` begins.
That's not replicated here so are we bypassing that logic then? The `always_check_provided_wallet_key` check there... I'm not sure if that's something necessary, need to refresh my memory and test out what it's doing.

Regardless, duplicating this function here in Innkeeper might be necessary for the pattern of preventing tokens for deleted tenants, but if anyone else has any thoughts/opinions can discuss here

plugins/traction_innkeeper/traction_innkeeper/v1_0/innkeeper/routes.py
```diff
@@ -522,6 +532,60 @@ async def tenant_create_token(request: web.BaseRequest):
...
+    wallet_id = request.match_info["wallet_id"]
+    wallet_key = None
+
+    mgr = context.inject(TenantManager)
+    profile = mgr.profile
+
+    # Tenants must always be fetch by their wallet id.
+    async with profile.session() as session:
+        rec = await TenantRecord.query_by_wallet_id(session, wallet_id)
+        LOGGER.warn("when creating token ", rec)
```
loneil, line 553: Warn probably should be down to debug

plugins/traction_innkeeper/traction_innkeeper/v1_0/innkeeper/routes.py
```diff
@@ -522,6 +532,60 @@ async def tenant_create_token(request: web.BaseRequest):
...
+    # into create_auth_token.
+
+    # if there's no body or the wallet_key is not in the body,
+    # or wallet_key is blank, return an error
+    if not request.body_exists:
+        raise web.HTTPUnauthorized(reason="Missing wallet_key")
+
+    body = await request.json()
+    wallet_key = body.get("wallet_key")
+    LOGGER.warn(f"wallet_key = {wallet_key}")
```
loneil, line 571: Would remove this entirely, it's `debug` in the Multitenant Provider, but logging the key in any situation, even debug, probably not good IMO (might want to remove that debug one in the other plugin)

plugins/traction_innkeeper/traction_innkeeper/v1_0/innkeeper/routes.py
```diff
@@ -522,6 +532,60 @@ async def tenant_create_token(request: web.BaseRequest):
...
+
+    # Tenants must always be fetch by their wallet id.
+    async with profile.session() as session:
+        rec = await TenantRecord.query_by_wallet_id(session, wallet_id)
+        LOGGER.warn("when creating token ", rec)
+        if rec.state == TenantRecord.STATE_DELETED:
+            raise web.HTTPUnauthorized(reason="Tenant is disabled")
+
+    # The rest is from https://github.com/hyperledger/aries-acapy-plugins/blob/main/multitenant_provider/multitenant_provider/v1_0/routes.py
+    LOGGER.warn(f"wallet_id = {wallet_id}")
```
loneil, line 558: Warn probably should be down to debug

plugins/traction_innkeeper/traction_innkeeper/v1_0/innkeeper/routes.py
```diff
@@ -1029,8 +1093,24 @@ async def register(app: web.Application):
                 "/multitenancy/reservations/{reservation_id}/check-in", tenant_checkin
             ),
             web.post("/multitenancy/tenant/{tenant_id}/token", tenant_create_token),
+            web.post(
+                "/multitenancy/wallet/{wallet_id}/token", tenant_wallet_create_token
+            ),
         ]
     )
+    for r in app.router.routes():
```
loneil, line 1101: can add a quick code comment here about what/why it's doing

plugins/traction_innkeeper/traction_innkeeper/v1_0/tenant/routes.py
```diff
@@ -109,6 +109,31 @@ async def setup_tenant_context(request: web.Request, handler):
     return await handler(request)
 
 
+def active_tenant_only(func):
```
loneil, line 112: Think this is unused?

loneil (changes requested):
Some minor code comments to start, but also want to see if anyone has any thoughts or things to look out for with the pattern of overriding the token route

plugins/traction_innkeeper/traction_innkeeper/v1_0/innkeeper/routes.py
esune: Off the top of my head, I think the duplication might be inevitable since we overriding multiple times the same function and there is - in theory - no guarantee that the plugins would be installed and executed in the same order. Which brings me to the follow-up question: would it cause issues to have the plugins installed in reverse order than what it was tested?

i.e.: currently we install multitenant provider and then innkeeper, with multitenant overriding built-in endpoints and innkeeper overriding the result of that. What happens if we reverse the order of installation? If we must install plugins in a specific order we may want to add a note in the docs about that.

Gavinok: We need to explicitly load multitenant before innkeeper for these overrides to function

Gavinok: I have now imported the original function used in the multitenant plugin inside of the innkeeper plugin to reduce the duplication of code
https://github.com/bcgov/traction/blob/block-api-for-deleted-tenant/plugins/traction_innkeeper/traction_innkeeper/v1_0/innkeeper/routes.py#L557

plugins/traction_innkeeper/traction_innkeeper/v1_0/innkeeper/models.py
```diff
@@ -357,12 +357,19 @@ async def soft_delete(self, session: ProfileSession):
         Soft delete the tenant record by setting its state to 'deleted'.
         Note: This method should be called on an instance of the TenantRecord.
         """
+        # Delete api records
+        recs = await TenantAuthenticationApiRecord.query_by_tenant_id(
+            session, self.tenant_id
+        )
+        for rec in recs:
```
jamshale, line 364: I don't mind this explicit for loop but just fyi you could use list comprehension here if you wanted.

plugins/traction_innkeeper/traction_innkeeper/v1_0/tenant/routes.py
```diff
@@ -109,6 +109,31 @@ async def setup_tenant_context(request: web.Request, handler):
...
+    @functools.wraps(func)
+    async def wrapper(request):
+        print("> innkeeper_only")
```
jamshale, line 115: Do you want this to be a print statement?

jamshale (commented):
Looking good to me 👍  I'll let the regular maintainers decide to approve.

plugins/traction_innkeeper/traction_innkeeper/v1_0/innkeeper/models.py
Gavinok: Is there a reason to use list comprehension here if we don't intend on using the list for anything? I may be misunderstanding

jamshale: Just if you want to write it in one line. You don't have to use the returned list. Sometimes it's more readable. But I don't know, people have different opinions.

plugins/traction_innkeeper/traction_innkeeper/v1_0/tenant/routes.py
Gavinok: Forgot to remove that on this branch. This code is now dropped

esune (commented):
Looks good to me, did a quick smoke test and it works as expected. 👍🏻

loneil (approved):
It's looking good to me, great to have it leverage the multitenant token function (rather than a full override) too

PR #1334, opened 2024-08-23
Dependency updates

loneil:
Update dependencies and things, remove some out-of-date things from readme


esune (approved):

esune:
The versions spread across package files make me wonder if we should try to find a way to centrally define/manage them - I always end-up forgetting to update them...

loneil:
> The versions spread across package files make me wonder if we should try to find a way to centrally define/manage them - I always end-up forgetting to update them...

Yes it's not ideal. At the least the display in the frontend should not be hard-coded (should take from package, or env)

Whether the "backend" of the Tenant UI and the frontend have independent versions... that _probably_ still makes sense as they are 2 separate node packages

Issue #1365, opened 2024-09-22
Edit's to make demo work on ARM architecture reliably


H4LL:
Previously the demo in /scriptswould fail when I tried to run it on ARM architecture. I made some edits to specify platform so that the docker environment knows to use virtualised hardware as no implmentation exits on native ARM. If it's of use to anyone I'd be happy to make a PR. Here's my branch for reference: https://github.com/H4LL/traction/tree/arm_virtualisation.

Thanks for creating traction. It's an awesome tool. I'm loving it so far!

swcurran:
Awesome and much appreciated if you could put in a PR.  Thanks!

PR #1371, opened 2024-09-24
Arm virtualisation

H4LL:
I made some updates to dockerfiles/ compose/ devcontainer files to specify amd64 as the platform. This was to ensure docker knew to virtualise hardware as as there is currently no implementation for ARM architecture.  It's been working well for me so I thought I would share in case it helps someone else. 

I have an issue open for this on #1365.


WadeBarnes:
@H4LL, Here is an example of a simpler pattern to accomplish the same thing.

https://github.com/bcgov/von-network/blob/e32d1732963ee64eb6635614f1ffb37db8b0cf36/manage#L8-L19

It instructs docker to use `linux/amd64` on ARM platforms without hard coding the use of `linux/amd64` throughout all of the docker and docker compose files.  I'm sure something similar can be done with the devcontainer to instruct it to use `linux/amd64` as well, but not sure the exact mechanics off the top of my head.

H4LL:
Ah, I see. Got it, I'll make some edits in due course to make it graceful like the von precedent, test and then try again. Thanks for the feedback!

H4LL:
@WadeBarnes Turns out it was super easy in the end and pretty much the command you passed me verbatim. I searched the von repo for any other reference to `DOCKER_DEFAULT_PLATFORM` but it turned up nothing so I suppose I just learned a handy environmental variable for docker. 

I tested the code by wiping my images and running a fresh build using the convenience script I've added. Everything is working out on my end. 
I think it may be getting into manage script territory though. If you'd like something along those lines, I'd be happy to put it together!

WadeBarnes:
That looks a lot cleaner.  Yes, I'd rather see a simple `./manage` script that can be used to build, start, and stop the services more easily.

Pro-Tip:
- Use `pushd` and `popd` to manage the directory context within the script.   This way you can change directories as needed throughout the script without inadvertently changing the working directory for the user.

scripts/README.md
```diff
@@ -147,6 +147,12 @@ This will leave the volume (data) intact and available on restart.
 docker compose down
 ```
 
+For mac users with ARM architecture, try using the conveniance script
```
esune, line 150: Nit.: convenience

esune (changes requested):
Thanks for the changes, looks a lot cleaner!

Some of your commits appear to not have DCO, could you please amend them to include it and push again?

H4LL:
Hey, I managed to make a hash of the amendments so I created a freshie here:
https://github.com/bcgov/traction/pull/1388

I hope it's what you were after!

PR #1372, opened 2024-09-24
Added middleware to prevent suspended tenants from making use of tokens

### LINKED ISSUE #1348: Delete existing tokens associated with the wallet of a tenant that has been soft deleted

Gavinok:
This PR resolves #1348 

In addition it also corrects the use of delete and deleted in the inkeeper UI


plugins/traction_innkeeper/traction_innkeeper/v1_0/tenant/routes.py
```diff
@@ -109,6 +111,36 @@ async def setup_tenant_context(request: web.Request, handler):
     return await handler(request)
 
 
+@web.middleware
+async def double_check_token(request: web.Request, handler):
```
esune, line 115: Nit.: might want to name `check_token_not_suspended` to be more explicit.

esune (commented):
Other than a small rename for clarity this looks good!

loneil:
Great stuff, works for immediately getting tenants locked out, tested some flipping suspended back and forth on tenant and innkeeper side.

One thing to think of for efficiencey maybe, the `setup_tenant_context` middleware above this new one also is called on every requerst with a tenant's token. (this one is introduced to facilitate tenant ID logging). It also fetches the Tenant record. So I think (though I don't know inner workings of `context` as well) we're now doing 2 DB fetches for tenant record on each API call. 
Is it possible what's needed is already in the context? Or the `setup_tenant_context` could be modified to store the tenant record so we don't need to introduce the second lookup?

esune:
@loneil I think you are right and it's two database calls. I am not 100% sure how to retrieve the tenant information across middlewares though, and unless we can guarantee that the setup tenant context middleware is executed before the token check (is it based on the order in which they are instantiated?) we may need to look up the tenant info again. Worth looking into though to reduce load.

Gavinok:
I figured the best thing to do was simply append the check to setup_tenant_context as we will already have access to the tenant record there. This massively simplifies the code and removes that performance penalty @loneil mentioned. I chose to leave the logging of the tenant record as is. Let me know what you think.

plugins/traction_innkeeper/traction_innkeeper/v1_0/tenant/routes.py
```diff
@@ -106,6 +111,10 @@ async def setup_tenant_context(request: web.Request, handler):
 
     log_records_inject(tenant_id)
 
+    # Ensure tokens are not associated with suspended tenants
+    if TenantRecord.STATE_DELETED == rec.state:
+        raise web.HTTPUnauthorized(reason="Tenant Is Suspended")
+
```
loneil, line 117: Seeing a 
```
File "/home/aries/.venv/lib/python3.9/site-packages/aiohttp_apispec/middlewares.py", line 45, in validation_middleware
return await handler(request)
File "/home/aries/traction_plugins/traction_innkeeper/v1_0/tenant/routes.py", line 115, in setup_tenant_context
if TenantRecord.STATE_DELETED == rec.state:
UnboundLocalError: local variable 'rec' referenced before assignment
```
on the deployed PR

loneil (commented):
Nice, I like that idea of simplifiying it to have the existing middleware handle all tenant context setup things (maybe if it gets really big or needs to be more specific for different routes could refactor later if that happens).

Think the deleted check needs to get moved into scope

loneil (approved):
Working all good for me, yeah I think that's the right spot for the rejection. Need a wallet coming out of the token (or else we don't care to check the state becasue it's not a tenant), and no need to set up the logging because we stop the call there. Caller gets a `401: Tenant Is Suspended` and the API shows that error too.

PR #1381, opened 2024-10-07
Helm Chart Improvements

i5okie:
- Do not create the database secret, use the secret created by PostgreSQL chart itself when enabled.
- Update Values file
- Update _helpers
- Update and clean up Readme

**Note:**
_When installing with a custom release name that does not contain "traction", the user must set `fullNameOverride`._


charts/traction/README.md
```diff
@@ -52,6 +52,12 @@ kubectl delete secret,pvc --selector "app.kubernetes.io/instance"=my-release
 
 ## Parameters
 
+| Name               | Description                                                                                                    | Value              |
+| ------------------ | -------------------------------------------------------------------------------------------------------------- | ------------------ |
+| `nameOverride`     | String to override the helm chart name, second part of the prefix                                              | '""'               |
+| `fullnameOverride` | String to fully override the helm chart name, full prefix. ***Must be set when using a custom release name.*** | '""'               |
```
esune, line 58: Should we be very specific and say "a custom release name that does not include the workd traction"?

esune (commented):
Small tweak suggestion.

Also wondering how will migration from existing deployments with the previous version of the chart work: if database users are created with different names we may risk issues when rolling out the new chart. Do we need special steps documented on how to update to this release?

charts/traction/README.md
i5okie: good point

i5okie:
> Small tweak suggestion.
> 
> Also wondering how will migration from existing deployments with the previous version of the chart work: if database users are created with different names we may risk issues when rolling out the new chart. Do we need special steps documented on how to update to this release?

There's no change to the database user names.
We are pointing to existing secrets in our deployments in dev, test, and prod.
I think PR re-deployments and current deployments will have PostgreSQL create its own secret, which will have to be manually updated with values from the existing secret.

esune (dismissed):

loneil:
@i5okie fairly standard regression testing of Traction stuff appears to be working as expected on the PR env created here.

loneil:
Are the changes with https://github.com/bcgov/traction/pull/1342 related or superseded by this one? Wondering if that older PR needs closing or merging.

i5okie:
> Are the changes with #1342 related or superseded by this one? Wondering if that older PR needs closing or merging.
superseded. I'll remove that pr.

esune (approved):

PR #1385, opened 2024-10-09
feat: Loki Log Streaming

amanji:


loneil:
Rebased and resolved package conflicts

loneil:
@amanji there's a pull request here that refactors some startup to a `manage` script. Any thoughts on if we merge that if that would cause any issues (or should have change to include) the docker compose startup instructions in this PR?

https://github.com/bcgov/traction/pull/1388/files

amanji:
@loneil we'd need to add this to the manage script:

https://github.com/bcgov/traction/blob/86fb4304e9c14f893551f8afb8e3abad7308e005/scripts/README.md?plain=1#L132

loneil:
FYI trying on the deployed PR and hitting start on the logs doesn't seem to show output or an error

See an aggregateError in the tenant UI backend logs:
image

amanji:
> FYI trying on the deployed PR and hitting start on the logs doesn't seem to show output or an error
> 
> See an aggregateError in the tenant UI backend logs: image

We need to build in three more services for the logging. I'm assuming the deployed services dont account for those yet

esune:
> > FYI trying on the deployed PR and hitting start on the logs doesn't seem to show output or an error
> > See an aggregateError in the tenant UI backend logs: image
> 
> We need to build in three more services for the logging. I'm assuming the deployed services dont account for those yet

@i5okie do you have any suggestiono n how to set things up for the supporting services? i.e.: we list them as pre-requisites or do we include (optional) depenedncies in the Helm chart?

i5okie:
> > > FYI trying on the deployed PR and hitting start on the logs doesn't seem to show output or an error
> > > See an aggregateError in the tenant UI backend logs: image
> > 
> > 
> > We need to build in three more services for the logging. I'm assuming the deployed services dont account for those yet
> 
> @i5okie do you have any suggestiono n how to set things up for the supporting services? i.e.: we list them as pre-requisites or do we include (optional) depenedncies in the Helm chart?

In my opinion, I think Traction should work without throwing exceptions or failing when it runs without Loki backend being configured. As such, I think Loki should be an optional component setup separately by the user. I'd imagine, if loki endpoint is not provided/log-streaming is disabled, the feature should be turned off in someway. Or at least say logs unavailable in a friendly (not app crashing-way)  manner on the page.

As for the `promtail`, it should be included as an optional sidecar.

scripts/docker-compose.yml
```diff
@@ -122,7 +134,9 @@ services:
         condition: service_started
     environment:
       - SERVER_TRACTION_URL=${SERVER_TRACTION_URL}
+      - SERVER_LOKI_URL=${SERVER_LOKI_URL}
       - FRONTEND_TENANT_PROXY_URL=${FRONTEND_TENANT_PROXY_URL}
+      - FRONTEND_LOG_STREAM_URL=${FRONTEND_LOG_STREAM_URL}
```
esune, line 139: These `envvars` need to be added to the chart as well

amanji: See [chart](charts/traction/templates/ui/configmap.yaml) if its correct.

amanji:
> > > > FYI trying on the deployed PR and hitting start on the logs doesn't seem to show output or an error
> > > > See an aggregateError in the tenant UI backend logs: image
> > > 
> > > 
> > > We need to build in three more services for the logging. I'm assuming the deployed services dont account for those yet
> > 
> > 
> > @i5okie do you have any suggestiono n how to set things up for the supporting services? i.e.: we list them as pre-requisites or do we include (optional) depenedncies in the Helm chart?
> 
> In my opinion, I think Traction should work without throwing exceptions or failing when it runs without Loki backend being configured. As such, I think Loki should be an optional component setup separately by the user. I'd imagine, if loki endpoint is not provided/log-streaming is disabled, the feature should be turned off in someway. Or at least say logs unavailable in a friendly (not app crashing) manner on the page.
> 
> As for the `promtail`, it should be included as an optional sidecar. We could put together documentation, with example configs on how to set Traction up with Loki. I can help with the latter two.

I'll update the PR to hide the log feature when the `FRONTEND_LOG_STREAM_URL` is not set.

esune:
> 
> I'll update the PR to hide the log feature when the `FRONTEND_LOG_STREAM_URL` is not set.

@amanji it does not look like changes have been pushed since this - can you confirm? Would you be able to push the above tweak so we can move forward with reviewing/merging the pr?

scripts/docker-compose.yml
amanji: @esune can you confirm if the included chart changes look correct for these variables?

loneil:
Getting a fail on one of the loki fields during Helm deployment

```
Error: template: traction/templates/ui/deployment.yaml:9:24: executing "traction/templates/ui/deployment.yaml" at <include (print $.Template.BasePath "/ui/configmap.yaml") .>: error calling include: template: traction/templates/ui/configmap.yaml:33:29: executing "traction/templates/ui/configmap.yaml" at <include "loki.fullname" .>: error calling include: template: no template "loki.fullname" associated with template "gotpl"
```

charts/traction/templates/ui/configmap.yaml
```diff
@@ -29,6 +30,7 @@ data:
   SERVER_SMTP_SECURE: {{ .Values.ui.smtp.secure | quote }}
   SERVER_SMTP_USER: {{ .Values.ui.smtp.user | quote }}
   SERVER_TRACTION_URL: http://{{ include "tenant_proxy.fullname" . }}:{{ .Values.tenant_proxy.service.port }}
+  SERVER_LOKI_URL: http://{{ include "loki.fullname" . }}:{{ .Values.loki.service.port }}
```
esune, line 33: `loki.fullname` will not be available unless Loki is deployed at the same time as Traction.
I think it might be best to just use a full URL from the values file in case Loki is not being deployed with the solution. Otherwise you'll have to add Loki as a dependent chart to Traction and handle on/off scenarios.

charts/traction/values.yaml
```diff
@@ -591,6 +591,18 @@ tenant_proxy:
...
+loki:
+  ## Loki service configuration
+  ##
+  service:
+    ## @param loki.service.type Kubernetes Service type
+    ##
+    type: ClusterIP
+    ## @param loki.service.port Port to expose for http services
+    port: 3100
+
```
esune, line 605: If we go with a "hard-coded" Loki url this should be able to go.

esune (changes requested):
A couple minor tweaks, looks good otherwise.

esune (approved):
LGTM 👍🏻 - Will log follow-up tickets as needed

amanji:
Thank you for taking this over @loneil

PR #1388, opened 2024-10-10
Manage script including arm virtualisation

H4LL:
I made a PR the other dayh about mac virtualisation: https://github.com/bcgov/traction/pull/1371#pullrequestreview-2352541904

I ended up turning it into a ./manage file. I hadn't originally included signoffs in my previous PR so I thought I'd amend it but I ended up touching a bunch of commits which were nothing to do with me. I thought it would be simpler to create a fresh PR here.

I tested this by removing all images then running through ./manage build, stop, start, restart.


WadeBarnes:
I'm assuming https://github.com/bcgov/traction/pull/1371 can be closed in favor of this PR.  Is that correct?

scripts/manage
```diff
@@ -0,0 +1,159 @@
...
+  compose_cmd=$(get_docker_compose_command)
+
+  # Check if the command is empty or not set correctly
+  if [ -z "$compose_cmd" ]; then
+    echo "Error: Failed to determine Docker Compose command." >&2
+    exit 1
+  fi
+
+  # Use the captured command to stop and remove containers
+  $compose_cmd down
```
WadeBarnes, line 97: `down` is a destructive operation which removes containers, volumes, and networks.  You should have a `stop` option which is non-destructive will stop the containers so they can be started again in the same state.

scripts/manage
```diff
@@ -0,0 +1,159 @@
...
+start_docker_containers() {
+  echo "Starting Docker containers..."
+    # Capture the Docker Compose command
+    compose_cmd=$(get_docker_compose_command)
+
+    # Check if the command is empty or not set correctly
+    if [ -z "$compose_cmd" ]; then
+      echo "Error: Failed to determine Docker Compose command." >&2
+      exit 1
+    fi
```
WadeBarnes, line 77: This code is repeated several times.  It would be better to setup and check `compose_cmd` once globally.

WadeBarnes (changes requested):

scripts/manage
H4LL: Thanks! this is fixed in [63fe016](https://github.com/bcgov/traction/pull/1388/commits/63fe01684cc60f44dba90f6174ed7eff6ff5b359)

scripts/manage
H4LL: Should be fixed in [31d6050](https://github.com/bcgov/traction/pull/1388/commits/31d6050b1c0422891693c5f1fcab70bf275fff4d)

H4LL:
Thanks for the feedback, I went ahead and closed that other PR. I've also added a UI to the manage script to make it a little nicer in https://github.com/bcgov/traction/pull/1388/commits/db33f52e235babd2f506d32a0389960d00eaeae6

WadeBarnes (commented):
Nice work!  In addition to the `stop`, I would recommend including a `down | rm` command to allow a user to reset their environment.  My previous comment regarding the the `down` and `stop` options was meant to recommend the addition of the `stop` command.

WadeBarnes (dismissed):

loneil:
This looks great, my one thing with the `./manage` scripts (which we do use in plenty of our projects so is a good idea to match in Traction here) is that if I'm on Windows and just pull the repo and follow the instructions nothing will happen so it can look like it's broken. I can use Git Bash or WSL. But with docker compose build/up I can just run in powershell/cmd.

Probably just best to cover that with some notes in the Readme, unless anyone else has other thoughts on using Unix-style scripts for starting up here?

esune (commented):
I agree with @loneil about providing a little bit more detail on the fact that the script needs to be run in a bash-compatible shell as it might not be obvious for everyone running the project for the first time. Other than that this looks great!

I'd rather have it added in this PR than a follow-up just so that the changes are self-contained (and since we already tweaked the instructions in the readme file).

loneil (commented):
Works for me, used Win11 and Git Bash, ran with ./manage no args and used prompts to build and start.

I usually check logs outside of the terminal I would use to start up so not a big difference to me, but noticed in the older `docker compose up` instructions you _would_ see the logs for all the containers tailing out, whereas with manage the way I ran it it sits like this

image

when we use the manage script on VCAuthN for example we have logs tail out

image

if anyone else has thoughts on if keeping that pattern is important?

H4LL:
Hey @loneil, 

I could explore a 'logs' option which could potentially overwrite the default behavior of the terminal for easier debugging. I'm sure it wont take too long (famous last words). If we want to go deep with that option I could also try a logs dir where the logs from each container are directed into their own file. I've not seen that design pattern in any other Aries adjacent repos but I feel like it'd be useful.

loneil:
> Hey @loneil,
> 
> I could explore a 'logs' option which could potentially overwrite the default behavior of the terminal for easier debugging. I'm sure it wont take too long (famous last words). If we want to go deep with that option I could also try a logs dir where the logs from each container are directed into their own file. I've not seen that design pattern in any other Aries adjacent repos but I feel like it'd be useful.

Hi @H4LL I think it's useful enough to get what we have here in and look at logging enhancement as a separate issue. I'll create one for that later and link it back. So going to approve this and merge it shortly unless anyone else has any thoughts.

Other ones we use on our team that just tail to the terminal are
https://github.com/bcgov/vc-authn-oidc/blob/main/docker/manage
https://github.com/hyperledger/aries-endorser-service/blob/main/docker/manage

but yeah I think there's some other patterns around other Aries apps as well.

loneil (approved):

loneil:
Issue to look into log output created: https://github.com/bcgov/traction/issues/1401

PR #1398, opened 2024-10-15
ACA-Py 1.0.1 update

loneil:
Use aca-py 1.0.1 image, plugins repo tag update

Not updating references to OWF yet, upcoming in 1.1.0

Update **Traction** display version to 1.0.0, figure next release which includes ACA-Py 1.x.x update we'd want to update major version? Otherwise could do 0.6.0


esune (approved):
👍🏻 

Regarding the TenantUI versioning scheme: we could follow the numbering ACA-Py uses in general (that is, also in the innkeeper plugin which we never update version-wise). I am however wondering if we should just version things independently since no real big change has been pushed to the application itself...

PR #1407, opened 2024-10-18
1.1.0 Upgrade and OWF reference changes

loneil:
Upgrades to Traction to switch image/repo/links/etc over to OWF, and upgrade to ACA-Py 1.0.0

Builds and deploys all good on OCP here on a PR, regression tested out
- Plugin usage
- Basic workflows 
  - create tenant, connect to endorser/DID reg, 
  - create schema, cred def, rev reg
  - connect to BC wallet, message, issue cred, verify, revoke cred
- imported existing schema 
- used Traction as holder and did connection/issuance between tenants

**Docs**

- Change "Hyperledger Aries Cloud Agent Python (ACA-Py)" references (and the like) to "OpenWallet Foundation ACA-Py"
- Github links update to OWF project

**TODOs/questions?**

There are still deployment paths locally and on Helm volume mount deployment that use `/home/aries`
Example `            mountPath: "/home/aries/ledgers.yml"` etc
My 2 cents is that if that is ok and doesn't cause any issues (doesn't appear to be here) then keeping "aries" as a legacy folder name on Traction isn't so bad? Better maybe than having a deployment risk of having to update things on existing deployments? Any thoughts?

On About page still have hyperledger Aries logo, should replace that with something when we have one
image

CODE_OF_CONDUCT.md is still a copy of hyperledger one, including "Hyperledger" references so will need to think about what that should be updated to. 

2 doc references to "Hyperledger Indy" is that still ok nomenclature?


jamshale (approved):
Looks good to me 👍 

For the .aries mountpath I think this was just a common name with dockerfiles. It doesn't have to be changed, but I think it could be without any problems.

jamshale:
The CODE_OF_CONDUCT.md might have the changes in aca-py now... I'm not sure. A few of these types of files are still being figured out. Can change it later.

Hyperledger Indy references are fine. Indy will remain a hyperledger project and traction is still using it exclusively for now.

esune (approved):
Looks good to me, thanks @loneil 

Regarding the `aries` folder mount: it looks like the user/home in the ACA-Py dockerfile is still `aries` so I'd say we keep it as-is for now and update things once/if the base image migrates to a different naming convention.

PR #1412, opened 2024-10-21
Readonly ledger support

### LINKED ISSUE #1296: Tenant UI: display read-only ledgers
Currently only ledgers for which an endorser exists are displayed in the Tenant UI, but we want to also display a list of ledgers connected in read-only mode that can be used to receive credentials and compose/validate proofs for issuers rooted on those.

Gavinok:
This PR resolves #1296


loneil (changes requested):
Not seeing a difference between the read-only candy-prod ledger and the ones with endorsers

image


I see the ledger settings have `is_write` and `read_only` which is confusing (maybe worth confirming what those 2 things are?).... but in our setups we're not able to connect or write to the candy-prod one so should be marking that one as read only, and hiding that connect button.

Let's have the column be something other than the pencil icon with a different color too (accessibility etc). Maybe a check/X for "writable" unless anyone has a better UI idea

services/tenant-ui/frontend/src/store/tenantStore.ts
```diff
@@ -27,6 +27,109 @@ import { useTokenStore } from './tokenStore';
 import { fetchItem } from './utils/fetchItem';
 import { fetchList } from './utils';
 
+interface ServerConfig {
```
loneil, line 30: Mind moving this to types\index.ts and importing from there? Otherwise looking all good

Gavinok: Sure thing

PR #1428, opened 2024-11-04
Helm chart update - Do not pre-create a custom database

i5okie:
Askar-Store is failing to provision a database if one already exists.
This PR addresses this issue by removing the `postgresql.auth.database` setting from the values. Which was used to trigger PostgreSQL sub-chart to pre-create the database.
- Remove `postgresql.auth.database` from Values

Clean up:
- Remove any provisions for `postgresql-ha`


esune (approved):

loneil (approved):
Looks good. Side note, would this affect a new PR deployment upon merge then? (I would try something out I was working on last week if so)

PR #1459, opened 2024-12-18
Remove unused endorser settings/parameters

esune:
Small cleanup removing duplicated settings for the endorser that were being set via environment variable AND configuration file.
Also removes a couple of settings (jsonld plugin and webhook endpoint) that are not required/used anymore.


loneil:
If unused (ENDORSER_ACAPY_READ_ONLY_MODE etc) could take them out of the `env-example` as well

loneil (approved):
Pulled locally, built from scratch works for me connecting to auto-endorser

PR #1483, opened 2025-01-18
Sass fixes and library updates

loneil:
Sass deprication warnings were throwing so much it was taking time to load Traction up. Quickly clean those up for better dev experience on other stuff.
Do some updates as well.

Simplified the toast import styles down as well (to make Sass update work better), but really I'd rather just use the included Primevue ones but that would require all the file changes so leave off for now.


services/tenant-ui/frontend/src/assets/layout.scss
```diff
@@ -1,4 +1,4 @@
-@import 'variables.scss';
+@use 'variables' as *; // Import variables as global
```
esune, line 1: Should this be `variables.scss` like in other locations?

loneil: Oh, yeah doesn't matter in Sass but should be consistent everywhere. Fixed

esune (approved):

PR #1489, opened 2025-01-31
[Breaking change] Update chart to use PostgreSQL 16.4

i5okie:
### Breaking Change
This PR updated the PostgreSQL to version 16.4.
Upgrading existing deployments will require manual intervention.


charts/traction/Chart.yaml
```diff
@@ -2,7 +2,7 @@ apiVersion: v2
 name: traction
 description: The Traction service allows organizations to verify, hold, and issue verifiable credentials. The Traction Tenant UI allows tenants to manage their agent.
 type: application
-version: 0.3.5
+version: 0.3.6
```
esune, line 5: I think `0.3.5` is actually the next version (it was already updated).

esune (changes requested):
Just a small tweak to the chart versioning. Also wondering if we should tag the PR as a breaking change due to the need of manual intervention to upgrade postgresql from lower versions to 16.

charts/traction/Chart.yaml
i5okie: Updated.

esune (approved):
Looks good to me. This will not affect our crunchy-backed clusters, so I think the only deployment that is affected is sandbox - which will get redeployed from scratch once the chart is published.

Other than a notice in the release notes I don;t think there is anything we need to do - correct?

PR #1504, opened 2025-02-11
Use 1.1.0 as Traction app version

loneil:
ACAPy has been updated so use a minor version bump


esune (changes requested):
Need to also update `Chart.yaml` to reflect the version change.

esune (approved):
👍🏻

Issue #1509, opened 2025-02-20
Fix invitation retrieval for peer DIDs in /connections/{conn_id}/invitation endpoint


MonolithicMonk:
**Description of the Issue:**

The `/connections/{conn_id}/invitation` endpoint in Traction (within the `traction_innkeeper` plugin) is intended to retrieve the invitation object associated with a connection. However, when invitations are created using `did:peer:2` or `did:peer:4` DID methods, the endpoint malfunctions and returns an error.

Specifically, the error is an `AttributeError: 'InvitationMessage' object has no attribute 'endpoint'`. This occurs because for peer DID methods, the `InvitationMessage` object omits the `endpoint` attribute (as it's not relevant for peer-to-peer DIDs). However, the current implementation of the `/connections/{conn_id}/invitation` endpoint incorrectly attempts to access `invitation.endpoint` when constructing the invitation URL. This leads to the error and prevents users from retrieving the invitation URL for peer DID invitations.

**Impact:**

Users are unable to reliably retrieve invitation URLs for  `did:peer:2` and `did:peer:4` invitations through the `/connections/{conn_id}/invitation` endpoint. This breaks expected functionality and hinders workflows that require retrieving invitations post-creation for these DID methods.

**Steps to Reproduce:**

1.  Create a connection invitation using the `/out-of-band/create-invitation` endpoint, specifying `use_did_method: did:peer:2` or `did:peer:4`.
2.  Note the `invi_msg_id` from the response and use it to retrieve the `connection_id` at the `/connections` endpoint.
3.  Attempt to retrieve the invitation using the `/connections/{conn_id}/invitation` endpoint, replacing `{conn_id}` with the noted connection ID.
4.  Observe the `AttributeError` in the server logs.

**Proposed Fix :**

1.  **Removing reliance on `invitation.endpoint`:**  The code no longer attempts to access the `invitation.endpoint` attribute.
2.  **Relying on `invitation.to_url(base_url)`:** The code utilizes `invitation.to_url(base_url)` to generate the invitation URL.
3.  **Handling Partial URLs from `to_url`:** The code now includes a check to detect if `invitation.to_url(base_url)` returns only query parameters (indicated by starting with `?`). If so, it prepends the `base_url` to construct the complete invitation URL.
4.  **Fallback for `base_url` Settings:** The code also includes a fallback mechanism to use `default_endpoint` if `invite_base_url` is not configured, and logs warnings if neither is available.

**Code Snippet of the Fix:**

```python
[...]

PR #1510, opened 2025-02-20
Fix: Correctly generate invitation URLs for peer DIDs in /connections…

MonolithicMonk:
…/{conn_id}/invitation endpoint fix to retrieve invitation urls for did:peer.  See [issue.](https://github.com/bcgov/traction/issues/1509)


loneil (approved):
Tested locally, can still fetch invitation with OOB invites and connections invitations (though those are deprecating), and also with did:peer2/4 after these changes.

PR #1511, opened 2025-02-24
feat(connections): Add alias field to invitation response schema

MonolithicMonk:
This PR adds the `alias` field to the invitation response schema

### Changes:
- Extended `InvitationResultSchema` to include an optional `alias` field.
- Added descriptions and examples for the `alias` field in the OpenAPI documentation.

### Why:
- This change allows tools like openapi-generator to generate the correct interfaces / classes for the `/connections/{conn_id}/invitation` endpoint

### Testing:
- Verified that the `alias` field is included in the response when fetching an invitation.
- Confirmed that the OpenAPI documentation reflects the new field.


loneil (approved):
Works for me, thanks for the PR.

PR #1558, opened 2025-04-01
Remove environment definitions from deployment jobs

loneil:
We removed the Environment Secrets in this repo, simplifying down to just using Repository Secrets.
The openshift token sitting in repository secrets it fell back to was out of date. Set that to the one we have in **TOOLS** labelled as "github-cicd-token"

image

Include Tenant UI minor version library updates as well.


loneil:
(secret precedence by the way)
If multiple levels of secrets are defined (e.g., repository, organization, and environment), the workflow will prioritize them as follows:
i.	Environment secrets (if the workflow specifies an environment).
ii.	Repository secrets.
iii.	Organization secrets.

esune (approved):

Issue #1406, opened 2024-10-16
Innkeeper plugin: normalize formatting/linting tools with other plugins


loneil:
The Innkeeper plugin has black and flake8 set up as dev dependencies (https://github.com/bcgov/traction/blob/main/plugins/traction_innkeeper/pyproject.toml#L21). In the plugin repo the plugins are setup to use Ruff instead. We should match build processes for this plugin to the ones in the repo.

Alter dev dependencies and format/lint setup to be like the plugins repo. Make sure any GHA build processes for PRs and merges and such use the same patterns where possible when it comes to lint checks as well.

The innkeeper plugin also needs unit testing additions and dev tools for that should match as well, but that's covered in https://github.com/bcgov/traction/issues/946

loneil:
Tagging for any additional thoughts @jamshale

jamshale:
It's not really that important if there's not going to be a lot of changes to this plugin... But, yes it would be good to be consistent with the other plugins.

I think some basic unit testing coverage is more important, and possibly some integration testing like the plugins have. It has caught issues with libraries that aren't going to be caught with innkeeper until manually tested in traction UI.

I think when someone adds some tests they could updating the linting/formatting/docs at the same time.

Gavinok:
Since I am already adding linting and testing for the plugin as a github action I may as well tackle this at the same time

PR #1565, opened 2025-04-04
Innkeeper unit tests and unify linters

Gavinok:
This PR resolves #702 and #1406


Gavinok:
We get lots of warnings which I can resolve but are existing problems with depreciated code in the main plugin. We should probably handle this in a seperate issue.

loneil (commented):
--

loneil (dismissed):
Nice, this is great. Tests are running on the PR deployment, great to be using the same dev tooling as the other plugins. Should definitely get this structure in and can adjust more code cleanup and add tests as time permits later on.

esune (approved):
👍🏻

PR #1547, opened 2025-03-27
[DON'T MERGE] Testing 1.3.0 RC

loneil:
ACA-Py update, and connections plugin use

PR #1578, opened 2025-04-16
ACA-Py 1.3.0 + WebVH Plugin

esune:
Testing deployment of the working version of the `webvh` plugin - see https://github.com/openwallet-foundation/acapy-plugins/pull/1654

### DIFF
--- a/plugins/docker/Dockerfile
+++ b/plugins/docker/Dockerfile
@@ -35,7 +35,7 @@ RUN pip install git+https://github.com/openwallet-foundation/acapy-plugins@1.3.0
 RUN pip install git+https://github.com/openwallet-foundation/acapy-plugins@1.3.0#subdirectory=multitenant_provider
 RUN pip install git+https://github.com/openwallet-foundation/acapy-plugins@1.3.0#subdirectory=rpc
 RUN pip install git+https://github.com/openwallet-foundation/acapy-plugins@1.3.0#subdirectory=connections
-RUN pip install git+https://github.com/openwallet-foundation/acapy-plugins@add-did-update@1.3.0#subdirectory=webvh
+RUN pip install git+https://github.com/openwallet-foundation/acapy-plugins@1.3.0#subdirectory=webvh
 
 
 ENTRYPOINT ["/bin/bash", "-c", "aca-py \"$@\"", "--"]
--- a/plugins/docker/Dockerfile
+++ b/plugins/docker/Dockerfile
@@ -1,4 +1,4 @@
-FROM ghcr.io/openwallet-foundation/acapy-agent:py3.12-1.3.0rc1 AS base
+FROM ghcr.io/openwallet-foundation/acapy-agent:py3.12-1.3.0 AS base
 
 # Install and Configure Poetry
 USER root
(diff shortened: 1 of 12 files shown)


plugins/docker/Dockerfile
```diff
@@ -34,6 +34,7 @@ RUN pip install git+https://github.com/openwallet-foundation/acapy-plugins@1.2.4
 RUN pip install git+https://github.com/openwallet-foundation/acapy-plugins@1.2.4#subdirectory=connection_update
 RUN pip install git+https://github.com/openwallet-foundation/acapy-plugins@1.2.4#subdirectory=multitenant_provider
 RUN pip install git+https://github.com/openwallet-foundation/acapy-plugins@1.2.4#subdirectory=rpc
+RUN pip install git+https://github.com/OpSecId/acapy-plugins@add-did-update#subdirectory=webvh
```
loneil, line 37: Note addition of connections plugin at https://github.com/bcgov/traction/pull/1547/files
`RUN pip install git+https://github.com/openwallet-foundation/acapy-plugins@1.2.4.1#subdirectory=connections`
this is to preserve 1.0 connections ability with the aca-py 1.3.0 release that removes them.

As well, if we're merging an RC, I don't think there's usually a plugin tag until the non-RC release is done, so normally we'd adjust all the plugins to the 1.3.0 tag instead of keeping 1.2.4, so we might need to verify if there's any necessary changes or anything (don't think so with the testing I did). Not sure if the webvh plugin here would be in the upcoming tag as well?

cc @esune

loneil:
For an aca-py version update we would also point the dependency for the Innkeeper plugin, can see here:
https://github.com/bcgov/traction/pull/1547/files#diff-851852fc27abbf85c6530a3340f5b0df10e4044716c81e94a1381339e69d44fb
(and then do a `poetry update` as well). Could probably do that as a separate PR as long as we don't immediately cut a version, or do it in this PR)

plugins/docker/Dockerfile
jamshale: This is merged to `acapy-plugins` now, so could be installed from there. After `1.3.0` is released we will create a plugins `1.3.0` release and then this will need to be updated again.

loneil: 1.3.0 tag is in now so this (and other plugin refs) can use that version now
https://github.com/openwallet-foundation/acapy-plugins/releases/tag/1.3.0

plugins/docker/default.yml
```diff
@@ -16,10 +16,12 @@ endpoint:
 # plugins
 plugin:
   - basicmessage_storage.v1_0
+  - connections.v1_0
```
jamshale, line 19: I think the connections plugin is just `connections`?

esune: I am confused about this the docs say it is just `connections`, however the plugin structure seems to indicate (as far as I can tell) the `v1_0` path/subfolder in the naming might be required?

jamshale: Ya. This was a bit confusing when we did it. The protocol should load in acapy because of the v1_0 directory but it didn't work correctly for some reason that isn't clear.

esune: Even using `conenctions` I don't see the right set of endpoints in swagger... 🤔

esune: There was a typo in the values file, should be fixed now... 😅

jamshale (approved):

PR #1643, opened 2025-06-23
Fix unique wallet name function

PatStLouis:
There was a recursive index appended to the wallet name resulting in names such as:
`alice`
`alice-1`
`alice-1-2`
`alice-1-2-3`

This fix will ensure only the current loop index is affixed to the wallet name so we get:
`alice`
`alice-1`
`alice-2`
`alice-3`


loneil (approved):
Looks good, get -1 -2 -3 now
image

PR #1644, opened 2025-06-25
Vite 7 tenent UI update

loneil:
Use Vite 7 in the frontend of the Tenant UI. Builds and runs all good for dist and dev.


Gavinok (approved):
LGTM

PR #1654, opened 2025-07-07
Tenant UI Backend: Transition to ESM, upgrade to ESLint 9

loneil:
2025 so make Node API fully ESM type (already using import/export everywhere etc but without type: module it resolves down to CommonJS). Adjust minor needs for that.

Switch to `tsx` for better ESM dev building.
Switch to Vitest as that plays far better with ESM and TS than Jest (and frontend already uses Vitest). Pretty much same syntax/API as Jest so not really any changes needed.

Upgrade Node API (not FE yet) to ESlint 9 with new flat file config. Add some prettier config that was always missing.


services/tenant-ui/frontend/package.json
```diff
@@ -7,8 +7,8 @@
     "dev": "vite",
     "build": "vue-tsc --noEmit && vite build",
     "preview": "vite preview",
-    "lint": "eslint --ext .ts,vue --max-warnings=0 .",
-    "lintfix": "eslint --fix --ext .ts,vue .",
+    "lint": "cross-env ESLINT_USE_FLAT_CONFIG=false eslint --ext .ts,vue --max-warnings=0 .",
+    "lintfix": "cross-env ESLINT_USE_FLAT_CONFIG=false eslint --fix --ext .ts,vue .",
```
loneil, line 11: Temp so FE can still find it's lint file without walking up the directory to find the BE flat file. (upgrade FE to eslint 9 later)

Gavinok (approved):

PR #1721, opened 2025-09-15
feat(chart): Update ACA-Py chart, disable wallet seed

i5okie:
The wallet seed in Traction is not utilized. The `acapy-0.2.2` helm chart adds the `.Values.secrets.seed.enabled` parameter, allowing the wallet seed to not be used.
- Bumped ACA-py chart dependency to take advantage of the latest changes
- Updated default values to set `.Values.acapy.secrets.seed=false`
- Updated documentation in values and README


esune (approved):
LGTM 👍🏻

PR #1812, opened 2026-01-16
Traction v1.3.0 release

PatStLouis:
## PR for a new release

Updates acapy to 1.5.0rc0
Removes all AIPv1 tests in favor of AIP2
Update to python 3.13
Address some dependabots vulnerabilities
Bump app version

## Plan for next version (v1.3.1)

Update acapy to 1.5.0
Update plugins to 1.5.0
Remove connections plugin (remove / update tests for did exchange)
Address bitnami common chart and publish new chart version

### DIFF
--- a/services/tenant-ui/frontend/src/types/index.ts
+++ b/services/tenant-ui/frontend/src/types/index.ts
@@ -37,8 +37,7 @@ export interface SchemaStorageRecord extends AcapySchemaStorageRecord {
   schema: Schema;
 }
 // The Types in the cred exchange format are just "object"
-export interface ExtendedV20CredExRecordByFormat
-  extends V20CredExRecordByFormat {
+export interface ExtendedV20CredExRecordByFormat extends V20CredExRecordByFormat {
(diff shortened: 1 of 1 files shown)


plugins/traction_innkeeper/poetry.lock
```diff
@@ -1,15 +1,15 @@
-# This file is automatically @generated by Poetry 2.2.1 and should not be changed by hand.
+# This file is automatically @generated by Poetry 2.0.1 and should not be changed by hand.
```
loneil: Might want to try with Poetry 2.2.1 or later to be sure of no inconsistency with downgrading

esune: The GH workflow specifies `2.1.1` as version of poetry to be installed, I think we should keep things in sync overall.

esune (commented):
We might want to do a pass and update dependencies before releasing a new version, since we have several dependabot-flagged upgrades we could complete.

PR #1813, opened 2026-01-21
Bump chart/app version and install issue credential v1 main

PatStLouis:
Add issue credential v1 and present proof v1 plugins

### DIFF
--- a/charts/traction/values.yaml
+++ b/charts/traction/values.yaml
@@ -159,19 +159,6 @@ acapy:
     wallet-storage-type: postgres_storage
     wallet-type: askar
     webhook-url: ""
-    plugin:
-    - --plugin 'aries_cloudagent.messaging.jsonld'
-    - --plugin traction_plugins.traction_innkeeper.v1_0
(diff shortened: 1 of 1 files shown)


charts/traction/values.yaml
```diff
@@ -159,6 +159,19 @@ acapy:
...
+    - --plugin-config-value traction_innkeeper.innkeeper_wallet.tenant_id=\"$(INNKEEPER_WALLET_TENANT_ID)\"
+    - --plugin-config-value traction_innkeeper.innkeeper_wallet.wallet_key=\"$(INNKEEPER_WALLET_WALLET_KEY)\"
+    - --plugin multitenant_provider.v1_0
+    - --plugin basicmessage_storage.v1_0
+    - --plugin connections
+    - --plugin connection_update.v1_0
+    - --plugin rpc.v1_0
+    - --plugin webvh
+    - --plugin present_proof.v1_0
+    - --plugin issue_credential.v1_0
```
esune: This seems to be duplicating the `extraArgs` section below (line 211). Which one of the two do we actually need?

esune (changes requested):

PR #1822, opened 2026-01-26
Fix revocation badge

PatStLouis:

PR #1852, opened 2026-02-25
chore(innkeeper): upgrade bcrypt to 5.x and add 72-byte compat

PatStLouis:
## Summary

Upgrades bcrypt from 4.2.1 to 5.x in the traction-innkeeper plugin and keeps the app working under bcrypt 5’s 72-byte password limit.

## Changes

- **bcrypt**: Bumped to `^5.0.0` in `plugins/traction_innkeeper/pyproject.toml`.
- **Bcrypt 5.x compatibility**: Added `bcrypt_compat.py` with `limit_for_bcrypt()` and use it in `utils.py` and `tenant_manager.py` so all inputs to `hashpw`/`checkpw` are capped at 72 bytes (bcrypt 5.x raises `ValueError` for longer inputs; 4.x truncated silently).
- **Plugin refs**: `multitenant-provider` and `connections` (acapy-plugins) are set to `rev = "main"` because 1.4.0 pins bcrypt 4.x and would conflict with bcrypt 5.x. Only these two plugins are on `main`; no other acapy-plugins were changed.
- **Tests**: New test `test_limit_for_bcrypt_long_password`; all 249 innkeeper tests pass with ruff lint/format and coverage.

## Note on `rev = "main"`

Using `main` is temporary until the acapy-plugins repo has a **release** that uses bcrypt ^5.0.0 for `multitenant_provider` and `connections`. Once that release exists, we should switch these dependencies back to a **tag** (e.g. `rev = "1.5.1"`) instead of `main`, so the innkeeper depends on a stable release instead of the current main branch.


plugins/traction_innkeeper/pyproject.toml
```diff
@@ -16,12 +16,13 @@ packages = [{include = "traction_innkeeper"}]
...
-bcrypt = "^4.2.1"
+bcrypt = "^5.0.0"
 mergedeep = "^1.3.4"
 typing-extensions = "4.15.0"
 anoncreds = "^0.2.3"
-multitenant-provider = {git = "https://github.com/openwallet-foundation/acapy-plugins", rev = "1.4.0", subdirectory = "multitenant_provider"}
-connections = {git = "https://github.com/openwallet-foundation/acapy-plugins", rev = "1.4.0", subdirectory = "connections"}
+# main: multitenant_provider and connections use bcrypt ^5.0.0
+multitenant-provider = {git = "https://github.com/openwallet-foundation/acapy-plugins", rev = "main", subdirectory = "multitenant_provider"}
+connections = {git = "https://github.com/openwallet-foundation/acapy-plugins", rev = "main", subdirectory = "connections"}
```
loneil: Can just swap these back to 1.5.1 tag (make sure we don't release Traction, not that it would probably cause problems though) when it's out

loneil (approved):
I switched the PR over to manual reservation approval and confirmed the generated password works all good. Tenant creation, Login and API keys work

plugins/traction_innkeeper/pyproject.toml
PatStLouis: yes, as soon as 1.5.1 is out, we will update these. I will rebase #1822 and address it there(that pr is waiting for 1.5.1 also). The work here is just so we can close the bcrypt issue. The multitenant plugin has no changes beside the bcrypt update

esune (approved):

PR #1929, opened 2026-04-08
Prepare 1.4.1 release

esune:
- Update versions to 1.4.1 where relevant
- Update dependencies (GHA, node, python)


loneil (approved):
More oddness with sandbox endorser but it's the same as before so no regression issue on the Traction/Tenant UI side

services/tenant-ui/frontend/tsconfig.json
```diff
@@ -3,7 +3,7 @@
     "target": "ESNext",
     "useDefineForClassFields": true,
     "module": "ESNext",
-    "moduleResolution": "Node",
+    "moduleResolution": "Bundler",
```
loneil, line 6: Oh good catch to do this, nver changed it when migrating to more modern TS way back.

PatStLouis (approved):
LGTM

PR #2017, opened 2026-07-03
Add "adopt wallet" tenant creation endpoint

loneil:
See https://github.com/bcgov/DITP-DevOps/issues/354

Adds a POST `/innkeeper/tenants/adopt endpoint` that creates a tenant record over an existing wallet (e.g. one relocated via acapy-tools **tenant-import**), with 404/409 guards for missing wallets and already-adopted ones. Also accepts an optional tenant name override through tenant creation.


jamshale (approved):
Makes sense to me 👍

esune (approved):

